PR #45, opened 2015-10-20
Use guava multimap

nboldi:


devaigergely81:
It saves us ~30 effective lines of code (plus corresponding tests), but adds a new dependency. The impact on the code using these libs is minor.
Questions:
- How stable was the MultiMap API of guava in the last couple of releases?
- What else could we use guava for, that would save us more code?
- How does this affect the development environment? (Think about many students who are working on this, and to keep the devenv simple for those who might fork and contribute to txtUML later.)

kmate:
As Eclipse Luna SR2 already contains guava, version 15.0.0.v201403281430. I think it's not a new dependency, when it is used in the IDE part. However, when used in runtime, it is a new dependency indeed.

nboldi:
The lack of some necessarry datastructures in the Java Standard Library led to the emergence of some very sophisticated (competiting) quasi-standard libraries.
One is guava, the other is apache commons. These libraries are stable.
I think that reinventing small parts of these implementations is bug-prone and unnecessarry.

devaigergely81:
I agree to use guava in the Eclipse-specific part, if it is already installed in Luna SR2. On the other hand, the JtxtUML API and runtime should have no external dependencies, because it should be possible to make an Eclipse-free release of those to be used in any Java dev env.

nboldi:
guava is not an Eclipse plugin. It is a library that can be found in Eclipse.

kmate:
It is not really a plug-in, but it has an OSGi packaging, which is very important when you want to depend on that. However, I agree with nboldi on the fact that sooner or later we could run into problems which could be solved a lot more easier with guava or commons. For example, the CLI interface of the model executor used apache commons to parse command line arguments. We had no problems to manage that single dependency at all.

nboldi:
We agreed on using a common library for using helpers that may be needed.

nboldi:
Please review: @gerazo

nboldi:
(This was just a test)

Issue #32, opened 2015-10-16
Action.create fails when the constructor has primitive parameters


nboldi:
The create action fails (returns null) when I give it a parameter of type int.
```A a = Action.create(PerformanceTestModel_.A.class, 10000);```

The constructor looks like this:
```
public A(final int numCycles) {
	//...
}
```

The workaround is to use `Integer` parameter instead of an `int` (but that is a validator error).

Also, when the `create` action fails it should raise an exception and stop the execution instead of silently returning a `null` that causes a `NullPointerException` later.

nboldi:
Are you currently working on this issue? If not I would take this.

nboldi:
I think an issue should not be closed until the fix is applied to trunk.

nboldi:
Merged now

Issue #18, opened 2015-10-15
txtUML Project should not be a plugin project


nboldi:
If this is used to handle dependencies a classpath container should be defined.

nboldi:
Are you currently working on this issue? If not I would take this.

kovacsgabor:
No, I am not working on it. Thank you if you take it.

PR #50, opened 2015-10-21
Primitive parameters fix

kovacsgabor:
Solved #32

### DIFF
--- a/examples/hu.elte.txtuml.examples.performance-test/src/PerformanceTestModel_.java
+++ b/examples/hu.elte.txtuml.examples.performance-test/src/PerformanceTestModel_.java
@@ -26,7 +26,7 @@ public class PerformanceTestModel_ extends Model {
 
 		int numForward;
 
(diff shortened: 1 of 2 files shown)


nboldi:
I planned to use guava's `Primitives` class here. No need to reinvent the wheel. Even a small wheel.

nboldi:
Of course it would mean that a new plugin dependency must be added to all project. Therefore I would wait for #18 with pulling this branch.

devaigergely81:
If the team accepts guava as a dependency, we can introduce it to solve this issue as well. However, I consider this issue important (almost all users will bump into this), so I would give green light to this pull request as is.

nboldi:
Ok. I think the whole `InstanceCreator` class needs to be refactored (it is a big spagetti), so this will be addressed then.

kovacsgabor:
I opened a new issue for further work on this (#65).

PR #52, opened 2015-10-21
Classpath container

nboldi:
Fix for #18. Uses copied code from Model Executor project. Strings are externalized.

Probably should be merged before #50.


kovacsgabor:
Apart from my two questions, it is good to go.

nboldi:
Fixed id name to `hu.elte.txtuml.project.runtimeClasspathInitializer`

nboldi:
Gábor, do you think it can be merged now?

PR #78, opened 2015-10-24
Fix/issue1

dobreffandras:
#1 fixed


devaigergely81:
I like the changes and have one more small request: Since these are user error messages, the
~~~~
+realException.getClass().getName()+ " thrown."
~~~~
part of the message could be omitted. The diagram name should not be in parentheses and the period (.) should be at the end of the first line.
arrangeerror

dobreffandras:
Should the diagram name stay in the sentence as is, or would one of the followig examples look better:

... of diagram: testpkg.testdiag.
... of diagram "testpkg.testdiag".
... of diagram 'testpkg.testdiag'.

devaigergely81:
I vote for the third option: ... of diagram 'testpkg.testdiag'.

devaigergely81:
Tested, works fine. Thank you.

PR #89, opened 2015-10-27
Fix resolution of partially qualified names

### LINKED ISSUE #25: Class name cannot be referenced in action code of state or transition
```
model PerformanceTestModel {
	class A {
		state S {
			entry {
				A a;
			}
		}
	}
}
```

The error message is the following:
```
Type mismatch: cannot convert the expression to Class	PerformanceTestModel.xtxtuml
```

djnemeth:
This is intended to fix #25.

By default in Xtext, references to local elements using partially qualified names are made possible with *local imports*. That is, when the hierarchical scope for a given context is being built, an implicit wildcard import statement is added for each nesting level. Given that the scope building process is correct (which fortunately is), the key aspect becomes the import resolution, which was the reason behind the mentioned problems indeed.

More precisely, the issue was caused mainly by the improper handling of Xtext's internal representation of qualified names during import resolution. In case of qualified names of nested types, Xtext uses the `.` character for language entities, and the `$` sign for their inferred `JvmType` counterparts to separate name segments. Consider the following example:

```
package p;

model M {
    class A;
}
```

Here `p.M.A` identifies the *language entity* corresponding to `A`, whereas `p.M$A` is the internal identifier of the *inferred Java class* (actually, only of its internal EMF representation).

Even though this solution can help distinguish between language and Java entities inside the framework, in our case it was the reason why partially qualified names couldn't be properly resolved. Technically speaking, some of these names *could be* resolved when certain conditions were met, but always to Java entities, resulting in type mismatches if language entites were expected instead.

To overcome the issue, I've redefined some aspects of the default import resolution process and adjusted the closely connected scoping components accordingly.

I hope I haven't broken anything.

### DIFF
--- a/examples/hu.elte.txtuml.examples.performance-test/src/ErrorReportingModel.xtxtuml
+++ b/examples/hu.elte.txtuml.examples.performance-test/src/ErrorReportingModel.xtxtuml
@@ -1,6 +1,0 @@
-
-model ErrorReportingModel {
-	class A {
-		
-	}
-}
--- a/examples/hu.elte.txtuml.examples.performance-test/src/PerformanceTestModel.xtxtuml
+++ b/examples/hu.elte.txtuml.examples.performance-test/src/PerformanceTestModel.xtxtuml
@@ -3,7 +3,7 @@ model PerformanceTestModel {
 	class Test {
 				
 		public void test() {
-			PerformanceTestModel.A a = create(A, 10000);
+			A a = create(A, 10000);
 			start(a);
 		}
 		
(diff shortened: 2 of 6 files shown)


djnemeth:
@nboldi and @devaigergely81, please review.

nboldi:
Seems working, @kmate, you could check the implementation.

kmate:
It would be reasonable to add unit tests for the new XtxtUMLMultimapBasedSelectable and XtxtUMLImportNormalizer classes.

kmate:
Something is wrong. When I try to get content assist between the inner braces of the following model, the `XtxtUMLImportNormalizer::deresolve` runs into and endless loop while trying to resolve `M` against imported namespace prefix `java.lang`.

~~~~
model M { class Test {

}}
~~~~

djnemeth:
On which expression do you call the content assist? You simply press Ctrl + Space with the cursor being in line 2 in your example?

kmate:
Yes.

Issue #6, opened 2015-10-14
Cannot install without Moka


devaigergely81:
Use clean Eclipse Luna Java edition, try to install the txtUML update site generated from txtUML trunk as of 2015-10-14.
The install fails, because it cannot find Moka.
~~~~
Cannot complete the install because one or more required items could not be found.
  Software being installed: txtUML 0.1.1.201510140920 (hu.elte.txtuml.feature.feature.group 0.1.1.201510140920)
  Missing requirement: txtUML Diagnostic Extensions 0.1.0.201510140920 (hu.elte.txtuml.diagnostics 0.1.0.201510140920) requires 'bundle org.eclipse.papyrus.moka.ui 0.0.0' but it could not be found
  Cannot satisfy dependency:
    From: txtUML 0.1.1.201510140920 (hu.elte.txtuml.feature.feature.group 0.1.1.201510140920)
    To: hu.elte.txtuml.diagnostics [0.1.0.201510140920]
~~~~

kovacsgabor:
Moka is needed because of txtUML visual debugging.

devaigergely81:
Yes, I know. But isn't it possible to configure the dependencies so that Moka is also installed automatically?

djnemeth:
We've found a [way](http://stackoverflow.com/a/8279204) to add update site references to the exported feature. I've tested the method on a clean Luna client, and it resolves the problem indeed, i.e. using a modified update site, the txtUML plugin can be installed automatically.

---

According to this solution, *after* the feature is exported, one has to modify the file `content.jar` – located in the root directory of the exported update site – by adding the required repository references to the packaged `content.xml`.

In our case, the following two entries have to be added to the `references` tag of the `repository` node:

```XML
<repository uri='http://download.eclipse.org/modeling/mdt/papyrus/updates/releases/luna/extra' url='http://download.eclipse.org/modeling/mdt/papyrus/updates/releases/luna/extra' type='0' options='1'/>
<repository uri='http://download.eclipse.org/modeling/mdt/papyrus/updates/releases/luna/extra' url='http://download.eclipse.org/modeling/mdt/papyrus/updates/releases/luna/extra' type='1' options='1'/>
```

As it appears to be undocumented, I specify the meaning of the used attributes:
- `uri`: the URI of the referenced repo
- `url`: the URL of the referenced repo
- `type`: whether the referenced repo is an artifact or metadata repo
[...]

PR #90, opened 2015-10-27
Configuring update sites correctly

gerazo:
Should fix https://github.com/ELTE-Soft/txtUML/issues/6

### DIFF
--- a/features/hu.elte.txtuml.feature/feature.xml
+++ b/features/hu.elte.txtuml.feature/feature.xml
@@ -107,7 +107,9 @@ This Agreement is governed by the laws of the State of New York and the intellec
    </license>
 
    <url>
-      <discovery label="Official txtUML website" url="http://txtuml.inf.elte.hu"/>
+      <update label="txtUML Update Site" url="http://txtuml.inf.elte.hu/releases/latest/"/>
+      <discovery label="XText Update Site" url="http://download.eclipse.org/modeling/tmf/xtext/updates/composite/releases"/>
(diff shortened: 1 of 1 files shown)


gerazo:
Please review @djnemeth @devaigergely81

djnemeth:
Unfortunately, I couldn't install the plugin automatically with a clean Luna. It still can't find the missing update sites:

```
Cannot complete the install because one or more required items could not be found.
  Software being installed: txtUML 0.1.1.201510271532 (hu.elte.txtuml.feature.feature.group 0.1.1.201510271532)
  Missing requirement: txtUML 0.1.1.201510271532 (hu.elte.txtuml.feature.feature.group 0.1.1.201510271532) requires 'org.eclipse.xtext 2.8.0' but it could not be found
```

However, the update sites specified in `feature.xml` have been added to the available software sites as I added the exported txtUML update site, but they are disabled by default:

update_sites

gerazo:
So even if we generate these lines into the content.xml as you have specified it is still not enough. I will take a look.

djnemeth:
I'm not sure if I have misunderstood your last comment, but the workaround I [proposed](https://github.com/ELTE-Soft/txtUML/issues/6#issuecomment-150872934) is about *manually* editing the `content.xml` *after* the plugin exportation is done. It's not very elegant indeed, but using that method I managed to install our plugin automatically.

gerazo:
You didn't misunderstand and you are right. It is not elegant and suffers from the same problem if the repositories are already added. Worst case we can try creating a composite site... I will try.

devaigergely81:
It fails also for me. If I manually enable the urls added in the first round, it installs everything automatically for the second try.

PR #91, opened 2015-10-27
Implement custom multiplicities in XtxtUML

djnemeth:
Custom association(end) multiplicities have already been made available in JtxtUML, provided by the generic `Multiple` class and annotations `Min` and `Max`. For additional details, see the [language guide](http://txtuml.inf.elte.hu/wiki/doku.php?id=v020:language#associations).

I have now implemented these custom multiplicities for XtxtUML as well. The syntax is as follows:
```
<multiplicity> ::= <int> ('..' (<int> | '*'))? | '*'
```

Some concrete examples:
- `*`
- `1`
- `1..4`
- `2..*`

As always, the multiplicity should be specified *after* the access modifier and *before* the class identifier in an association end. For example:
```
association L {
    public 1..4 A a;
    public 2..* B b;
}
```


djnemeth:
@kovacsgabor, please review.

devaigergely81:
Reviewed, tested, seems OK.

PR #93, opened 2015-10-28
3rd party update sites

gerazo:
Temporary solution needing some hand editing until releng is ready.

### DIFF
--- a/features/hu.elte.txtuml.feature/feature.xml
+++ b/features/hu.elte.txtuml.feature/feature.xml
@@ -107,7 +107,9 @@ This Agreement is governed by the laws of the State of New York and the intellec
    </license>
 
    <url>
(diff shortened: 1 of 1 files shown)


gerazo:
So what do you think @djnemeth @devaigergely81

djnemeth:
I'd like to have one tiny remark: *XText* should be written as *Xtext* (the second letter being lowercase) in the label of the update site reference at [line 111](https://github.com/ELTE-Soft/txtUML/blob/4317c671b8e290fdeb29767ae0d2961eef1ff214/features/hu.elte.txtuml.feature/feature.xml#L111) of `feature.xml`.

Apart from that, I agree that we can use this temporary solution until releng is ready. :shipit:

gerazo:
Thanks, I will do as you suggested

Issue #98, opened 2015-10-30
Transitions not allowed in composite states


devaigergely81:
The Xtext grammar does not allow transitions to be in composite states. The following gives an error:
~~~
model TestModel {
	class TestClass {
		initial Init;
		composite C {
			initial CInit;
			state A;
			transition CInitial {
				from CInit;
				to A;
			}
		}
		transition Initial {
			from Init;
			to C;
		}
	}
}
~~~
It is possible to write the transitions of the composite state outside, using fully qualified state names, but the initial transition is not triggered that way:
~~~
model TestModel {
	class TestClass {
		initial Init;
		composite C {
			initial CInit;
			state A;
		}
		transition Initial {
			from Init;
			to C;
		}
		transition CInitial {
			from testpkg.TestModel.TestClass.C.CInit;
			to testpkg.TestModel.TestClass.C.A;
		}
	}
}
~~~
~~~
TestClass:obj_1 leaves initial_pseudostate:Init
TestClass:obj_1 uses transition:Initial (initial_pseudostate:Init --> composite_state:C)
TestClass:obj_1 enters composite_state:C
TestClass:obj_1 enters initial_pseudostate:CInit
~~~
The missing step is CInit to A at the end.

PR #99, opened 2015-10-30
Fix/issue98 crlf

devaigergely81:
This is intended to solve #98.
* Most of the changes are in generated files by Xtext. These are currently in the repos, therefore I decided to commit their changes. When we will have proper build environment, we might decide removing these generated sources.
* The XtxtUML-related part of the repos has CRLF line endings (including all the Xtext-generated files). I kept this encoding for now. (Actually, most of the commits in this branch is about restoring CRLF.) Later we should decide how to unify line endings in the repos and elaborate the necessary settings on developer machines.
* The changes in non-generated files are:
  * `XtxtUML.xtext`: The grammar has been changed to support transitions in states.
  * `XtxtUMLValidator.xtend`: Validation for duplicate transitions needed to be changed accordingly.
  * A `.gitignore` file has been added in the `hu.elte.txtuml.utils.tests` project in order to make git put the src directory under version control. This eliminates a build error in this project, and is independent of #98.


devaigergely81:
@kmate and/or @djnemeth, please review!

djnemeth:
Seems to be working. I have tested the grammar, the previously improper inner transition execution and the adjusted duplicate name validation.

It's nice to see how we hadn't had to modify the model inferrer, thanks to the use of Xtend's dispatch methods.

kmate:
I think we should not merge back this branch, as it corrupts the history of the whole xtext language definition file. Please apply these changes on a fresh branch eg. using a patch.

devaigergely81:
OK, I'll redo this on another branch.

PR #101, opened 2015-11-03
Fix/issu98 crlf2

### LINKED ISSUE #98: Transitions not allowed in composite states

devaigergely81:
This should fix #98.
The CRLF problem seems to be avoided this time.
Most of the changed files are generated. The handmade ones:
 * XtxtUML.xtext
 * XtxtUMLValidator.xtend
 * hu.elte.txtuml.utils.tests/src/.gitignore (added to solve the missing src directory problem)


devaigergely81:
@kmate, please review

kmate:
It seems ok, but again, where are the tests? :D

PR #108, opened 2015-11-04
Signal-to-deleted-object warning now prints the signal (using its overridable toString method).

### LINKED ISSUE #60: Show signal type in signal-to-deleted-object warning
When running a JtxtUML model, if an object is deleted, and then it would receive a signal, the message ```"Warning: signal arrived to deleted model object TestModelClass:obj_1"``` is shown.

In this message, the type of the signal should be shown.

It was also proposed to display the data carried within the signal. However, as signals (optionally) carry data within them as fields of the representing Java class, I think it would be very slow and not very informative to list them with Java reflection.

kovacsgabor:
Fixed #60.


kovacsgabor:
@nboldi, please review.

nboldi:
I do not see any problem with that change.

Issue #16, opened 2015-10-15
Reduce reflection in txtUML runtime lib


devaigergely81:
Goal: speed up execution.

devaigergely81:
The reason for lowering the prio is that Boldi's recent measurements show that txtUML runtime performance is good as is.

PR #110, opened 2015-11-04
Collections can now contain any type of elements (including primitives); linking and unlinking optimized.

kovacsgabor:
Implemented as part of #16.


gerazo:
Reviewed. Let's hit master!

PR #119, opened 2015-11-07
Fix name deresolution in XtxtUMLImportNormalizer

### LINKED ISSUE #25: Class name cannot be referenced in action code of state or transition

djnemeth:
The previous implementation of qualified name deresolution caused an infinite loop if the imported namespace wasn't a semantical prefix of the to-be-deresolved name. This is now fixed by an initial check, which determines if custom deresolvation is necessary.

This is intended to fix #25.
For additional information, see #89.


djnemeth:
@kmate, please review.

kmate:
I think we can improve the performance of these code snippets, but I will do that in the xtxtuml-speedup branch. I will also add unit tests for these classes.

Issue #84, opened 2015-10-25
State machine diagram check box not persisted


devaigergely81:
The Papyrus model creation wizard very well performs almost all of its data, so that the values are the same when one reopens the wizard. The exception is the checkbox about state machine diagrams: That gets checked whenever the wizard is reopened. Please persist that one as well.

PR #122, opened 2015-11-09
Persisting "generate State Machines automatically"

dobreffandras:
Fixed issue #84


kovacsgabor:
:+1: The fix works (tested it in a runtime Eclipse).

PR #125, opened 2015-11-10
Instead of interruption, an explicit check is performed in every iteration of ModelExecutorThread's loop

### LINKED ISSUE #112: Refactor thread handling
`Thread.interrupt()` should not be used as it may never actually interrupt the thread if it gets to work continously: https://docs.oracle.com/javase/tutorial/essential/concurrency/interrupt.html

kovacsgabor:
To fix #112.


kovacsgabor:
@nboldi, please review.

nboldi:
I think this would not work, since the runtime can be in a waiting state when calling inactivate. I already have another solution for this problem.

nboldi:
See #127

kovacsgabor:
> I think this would not work, since the runtime can be in a waiting state when calling inactivate. I already have another solution for this problem.

That is why I call interrupt() in inactivate.

nboldi:
I see. The description suggested that this solution does not use interrupt, so I wasn't looking for it.

kovacsgabor:
OK. Nevertheless, this fix is unnecessary as you implemented your one.

PR #127, opened 2015-11-10
Refactored shutdown and shutdownNow, without using interrupts.

nboldi:


nboldi:
To fix #112

nboldi:
Tests are not updated yet.

plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/ModelExecutorThread.java
```diff
@@ -76,24 +79,20 @@ private void newMailboxEntry(MailboxEntry entry) {
...
-			mailbox.put(new ShutdownAction());
-		} catch (InterruptedException e) {
-		}
+	/**
+	 * Sets the model executor to be shut down immediately. If the execution of
+	 * a {@link #mailbox} item does not terminate it cannot guarantee the
+	 * termination.
+	 */
+	void shutdownImmediately() {
+		shutdownImmediately = true;
```
kovacsgabor: This method can be called from other threads. Changing a field without any synchronization is unsafe.

plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/ModelExecutorThread.java
```diff
@@ -76,24 +79,20 @@ private void newMailboxEntry(MailboxEntry entry) {
...
-						ModelExecutor.shutdownNow();
-					} else {
-						try {
-							mailbox.put(new ShutdownAction());
-						} catch (InterruptedException e) {
-						}
-					}
-				}
-			}
+		shutdownWhenEmpty = true;
```
kovacsgabor: This method can be called from other threads. Changing a field without any synchronization is unsafe.

plugins/hu.elte.txtuml.utils/src/hu/elte/txtuml/utils/NotifierOfTermination.java
```diff
@@ -8,10 +8,6 @@
 	private final List<Runnable> list = new ArrayList<>();
 	private boolean terminated = false;
 
-	public synchronized boolean isTerminated() {
-		return terminated;
-	}
```
kovacsgabor: Why change a utility class?

plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/ModelExecutor.java
```diff
@@ -66,6 +60,7 @@
 	 * an uninstantiatable class.
 	 */
 	private ModelExecutor() {
+		thread.start();
```
kovacsgabor: This code never actually runs.

See the JavaDoc:
"Sole constructor of <code>ModelExecutor</code>, which is designed to be
an ***uninstantiatable*** class."

kovacsgabor: Fixed in last commit.

kovacsgabor:
The newly created test still fails with java.lang.IllegalThreadStateException. I did not look further into the matter. (Another test also failed, I opened a new issue about this: #133.)

plugins/hu.elte.txtuml.utils/src/hu/elte/txtuml/utils/NotifierOfTermination.java
nboldi: It behaves like a notifier, an event source, so it should only communicate that way.

plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/ModelExecutorThread.java
nboldi: I don't think there could be synchronization problems since the method only sets the flag to true. There is no method to set this flag to anything except true.

nboldi:
Funny, the new test is green for me.

nboldi:
Oh, sorry, your new test.

nboldi:
All api.model tests are green except the `Action.create` problem mentioned in #133. Written some additional tests to completely cover the `ModelExecutorThread` class.

Note that `shutdownNow()` does not shut down the model if it has an infinite loop in the action code. I think there is no conventional way to do that.

plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/ModelExecutorThread.java
kovacsgabor: If the ModelExecutorThread instance reads this variable, it can cache its value and reuse that cached value unless some kind of synchronization happens. If another thread changes the value of this variable, it may not affect the ModelExecutorThread in any way.

plugins/hu.elte.txtuml.utils/src/hu/elte/txtuml/utils/NotifierOfTermination.java
kovacsgabor: I see you point. I agree.

kovacsgabor:
No, it was your test that failed for me. Only you did not include it in AllTests. That was the only change I did to the tests.

But yes, now everything seems fine about the tests.

My only remaining concern is then about the synchronization of the two boolean flags.

plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/ModelExecutorThread.java
nboldi: I think there is no problem because only the model executor thread reads it. Change will not appear immediately in that thread, but unsynchronized execution of shutdown cannot resolve in infinite loop (cannot set accidentally the already true flag to false).

nboldi:
@devaigergely81, I introduced a dependency here, that should be handled by maven after release engeneering is done, but currently just 3 jar files committed into the repository.

nboldi:
Furthermore please notice that `shutdownNow` does not wait for shutdown. In the original version shutdown is more likely to stop the thread before returning, so in the tests it was a conception that the method stops the thread (and executes the shutdown events) before exiting.

plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/ModelExecutorThread.java
kovacsgabor: But it may never notice that flag was set to true.

Furthermore, if it is waiting for the mailbox to have an entry, you wake it by giving it an entry. However, if it does not update the value of the flag and uses the cached value, then it enters the next iteration and waits again for the mailbox to get an entry. And will wait forever. Or at least, until you kill the JVM.

nboldi: I see your point. Setting the flags volatile should solve it, right?

kovacsgabor: Either that or using AtomicBoolean instead. I do not know which one is faster.

nboldi:
OK to merge, @kovacsgabor?

kovacsgabor:
Yes.

nboldi:
@devaigergely81, please review the changes in testing

nboldi:
Resolved the mock library using powermock-osgi bundle. It can be found on this update site: http://powermock-osgi.googlecode.com/svn/updateSite/1.5.6.0

gerazo:
Boldi, do we need to change something related to mockito dependency? Or is it final that you have committed it into the repo?

nboldi:
It is the final version. But the powermock-osgi bundle have to be included in the development environment, as well as in the test target.

gerazo:
Ok. I see that otherwise this branch can go. I add the dependency and will probably merge it as well.

Issue #133, opened 2015-11-11
A test of api.model fails


kovacsgabor:
The test about calling Action.create with wrong parameters fail with IllegalArgumentException.

gerazo:
These tests were disabled in maven, see https://github.com/ELTE-Soft/txtUML/commit/6b896f0f59e00500a4a0521f13eb7d1d7bc2eff2
If you are ready with the fix, do not forget to remove the pom.xml changes to put the tests back alive.

Issue #131, opened 2015-11-11
The runtime API lacks proper error handling


kovacsgabor:
The current implementation of the runtime API is based on the following rules:
 * If the model is invalid, nothing is guaranteed.
 * After reporting an error, nothing is guaranteed.

By "nothing is guaranteed", I mean that in these cases the API may throw surprising or uninformative exceptions (typically NullPointerExceptions) or result in an unspecified behavior.

The team agreed that a better error handling is required.

nboldi:
My proposal:
 - The runtime can assume a valid model and should not manually check model validation. If it is clear that the model is invalid, it should exit with a general error message. (Validation is improving.)
 - If the user performs an operation that is illegal in the current state, the runtime should roll back the changes and the execution should continue as if nothing happened. If it is not possible (for example: the operation had some return value that could be used later), it should stop after giving the error message to the user.

PR #134, opened 2015-11-11
Action.create does not throw IllegalArgumentException.

kovacsgabor:
Quick fix for #133.

Note:
Returning null will cause a NullPointerException in almost all cases, but it follows the original approach (to report the error and return null). In the long run, the error handling of the whole runtime will need a refactoring (see #131 ).

### DIFF
--- a/plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/Action.java
+++ b/plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/Action.java
@@ -3,6 +3,7 @@ package hu.elte.txtuml.api.model;
 import hu.elte.txtuml.api.model.ModelExecutor.Report;
 import hu.elte.txtuml.api.model.backend.MultiplicityException;
 import hu.elte.txtuml.utils.InstanceCreator;
+import hu.elte.txtuml.utils.RuntimeInvocationTargetException;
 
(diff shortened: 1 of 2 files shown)


kovacsgabor:
@nboldi, please review.

nboldi:
`RuntimeInvocationTargetException` means that an exception happened while executing the constructor. Shouldn't we throw the exception in that case?

kovacsgabor:
You mean we should throw the ```RuntimeInvocationTargetException``` or the contained exception itself? Because, in the latter case, Action.create would be declared to throw any kind of exception.

nboldi:
What is the current policy on declared exceptions in the model?

kovacsgabor:
Currently exception handling is completely forbidden in the model.

nboldi:
Why `RuntimeInvocationTargetException` is needed? `tryCreateWithConstructor` could simply throw an `InvocationTargetException` that is caught in `create`.

kovacsgabor:
Because we use ```InstanceCreator.create``` more than once and it seemed easier to use a ```RuntimeException``` in cases when we instantiate classes the constructors of which are known to throw no exceptions.

devaigergely81:
@nboldi, @kovacsgabor, have you resolved the above discussion on the exception type?

nboldi:
I think its OK to merge.

Issue #120, opened 2015-11-08
hu.elte.txtuml.layout.visualizer.helpers


kovacsgabor:
Some utility features are implemented in the ```hu.elte.txtuml.layout.visualizer.helpers``` package unnecessarily. I found the following two examples:

 * A ```Pair``` implementation is already used by multiple other txtUML projects, and is defined in ```hu.elte.txtuml.utils```.
 * A ```BiMap``` implementation is present in the ```com.google.common.collect``` package, a part of Guava, which we already have a dependency on.

Also, it should be considered to move some other general-purpose classes of this ```helpers``` package (like ```Triple```, ```Quadraple```) to the ```utils```project to let other txtUML projects use them in the future.

kovacsgabor:
Another minor issue regarding this package: The JavaDoc of ```Quadraple``` is wrong (it is the same as the documentation of ```Triple```).

GregoBalu:
I did everything suggested above except BiMap.

kovacsgabor:
Is there a specific reason you ignored my suggestion about BiMap? If there is, please explain it.

kovacsgabor:
(Unless there is a good reason for it, we agreed that we should not reinvent the wheel. That is why I suggested using the existing BiMap implementation.)

PR #135, opened 2015-11-11
Fixed issue #120

GregoBalu:
Issue #120 :
Removed Pair, Triple, Quadruple from layout.vis.helpers and added Triple and Quadruple to txtuml.utils.
Refactored Triple and Quadruple to mathc the pattern of Pair.
Refactored layout.visualizer project to use txtuml.utils.


kovacsgabor:
I added hashCode and equals methods to both Triple and Quadruple. I also fixed the same methods of Pair the original implementation of which could not manage null values.

Please pull with --rebase option to include my changes, and review them

kovacsgabor:
I see that the ToFirstPair, ToSecondPair and ToThirdPair methods of both Triple and Quadruple are unused. Also, they do not seem to be very helpful. (Furthermore, they do not obey the Java convention to begin with a lowercase letter, but this could be easily fixed.)

Do we need these?

PR #137, opened 2015-11-11
Replaced InnerClassInstancesMap with ClassToInstanceMap of Guava

kovacsgabor:
(Had no issue about it, just an improvement.)


kovacsgabor:
@nboldi, please review.

nboldi:
I don't really understand why is this needed, but I agree with the replacement of our own implementation with a more standard one.

kovacsgabor:
This way, I could remove a SuppressWarning about an unchecked cast. Also, it simplifies the API as we have less code without any drawbacks.

PR #142, opened 2015-11-16
Release engineering part1 mavenize, mavenize, mavenize

gerazo:
- Maven is now in charge with all its basic settings.
- Eclipse still works as it should.
- The target project is already there and can be used to configure target platforms correctly (it is still not using mirrors, so a bit slow)
- api tests package configuration was corrected
- "project" project was moved to plugins
- devenv document is already update to reflect these changes: https://github.com/ELTE-Soft/txtUML/wiki/Installation-of-the-Development-Environment


gerazo:
@kmate @kovacsgabor @devaigergely81 Can you take a look?

kmate:
I suggest to format all pom.xml files with Ctrl+Shift+F, as currently they are containing both tabs and spaces. Everything seems to compile, but some tests are failing.

gerazo:
xmls are formatted now. The tests will be disabled until they are not fixed. I will not do this in thiss pr as @kovacsgabor will fix a couple very soon.

PR #162, opened 2015-11-27
Version bump 030

gerazo:
Version is bumped to 0.3.0 and variable is used in pom.xmls so we shouldn't edit them by hand anymore


gerazo:
As it turns out, variables work everywhere except in version tag. And this is intentional. :-1:

gerazo:
Please review @kmate

kmate:
Seems ok, every dependencies are updated and the build succeeded, so I confirm the merge.

PR #171, opened 2015-12-01
Split model wizards

nboldi:


nboldi:
@devaigergely81, please test.
Container and package is guessed from selection. Please also validate that.

devaigergely81:
I suggest to fix #175 before merging. Other than that, I think it is all right.

nboldi:
Note that I removed the {txtUML} image from the wizards because it slided out of the view when a validation error appeared.

kovacsgabor:
The wizard only lets valid Java identifiers as model names but txtUML models may now have any name that is allowed in EMF-UML2.

kovacsgabor:
If package ```x.y``` is a model, the wizard still allows me to create a model 'above' it in package ```x```.

kovacsgabor:
Only a proposal, but I would prefer the XtxtUML syntax to be default.

kovacsgabor:
I created an issue that is connected to this branch: #182.

devaigergely81:
We have to think about it, but having a model in another model (eg. both in packages x and x.y) might not be a problem.

devaigergely81:
#175 is now solved on this branch. I will merge this and create issues about kovacsgabor's suggestions above.

PR #172, opened 2015-12-01
Normalize line endings

gerazo:
Line settings are controlled in .gitattributes file: everyone will have their native (LF everywhere, CRLF on Win) line endings when checking out but only LF will be stored in the repo.
All incorrect line endings are changed to LF. (commit bomb)


gerazo:
The process was roughly the following: https://help.github.com/articles/dealing-with-line-endings/

gerazo:
Please review @devaigergely81 , the only important changes is the .gitattributes file which should work even when someone doesn't set core.autocrlf

gerazo:
This last merge from master was horrific. Please review!

PR #178, opened 2015-12-02
Add jacoco maven plugin

gerazo:
It should do coverage reports on all functioning test projects. We will add the pickup of data to Jenkins as well.


gerazo:
Can you @kmate it please?

Issue #186, opened 2015-12-03
Unversion Xtext-generated code


devaigergely81:
Originally, we decided to put the generated code of Xtext under version control to make the development environment simpler. Now that we have Maven-based releng in place: Is it possible to do the 'Generate Language Infrastructure' step automatically, so that we could un-version the generated files? If that cannot be part of the build, we might still consider un-versioning and adding one more manual step to the setup of the development environment. What do you think, @ELTE-Soft/all?

gerazo:
We should unversion them at all costs. Even if it needs extra setup steps. I will look into it.

gerazo:
Done

PR #199, opened 2015-12-11
Generate Xtext also automatically

gerazo:
1. Every generated content is out of the repo.
2. Maven generates Xtext related stuff automatically in the build process.
3. The exact same stuff is generated as using Eclipse.
4. Manifest modification is disabled so git does not complain about modified manifests all the time
(In case of Xtext/Xtend version upgrade /which would be beneficial now I think/, the manifests can be merged by hand as the generated one is put next to the original.)
5. In case of a fresh checkout in Eclipse you still need to manually execute the "Generate Language Infrastructure" or do a maven build launch


gerazo:
Please review @djnemeth @kmate @devaigergely81

gerazo:
Related issue: https://github.com/ELTE-Soft/txtUML/issues/186

gerazo:
You can also review @nboldi :)

nboldi:
I could not find the `Generate Language Infrastructure` launch, and `Maven generate-sources` was not successful, but after a `clean-verify` and some refreshing everything worked fine. I suggest that doing a `clean-verify` should be the standard procedure for generating the sources.

So in general, it seems to be fine.

PR #201, opened 2015-12-14
Restructuring

nboldi:
txtuml_arch 1


nboldi:
@kovacsgabor review please

kovacsgabor:
I think hu.elte.txtuml.utils should remain platform-independent as the API has a dependency on it. Therefore, I would suggest moving the Eclipse-related utility classes from this project to another one. (A new one or the hu.elte.txtuml.utils.jdt with along with a rename of the project.)

gerazo:
This can be the separation where normal logging resides in one package while the one (an implementation) capable of interacting with the plugin logs is in the other.

nboldi:
Did the last four commit fix your issues?

kovacsgabor:
hu.elte.txtuml.api.model.tests using Mockito are not deterministic. Sometimes they fail, sometimes they do not.

kovacsgabor:
Please, review my updates. I think the branch is ready to be merged into master.

gerazo:
Were the flaky mockito test problems resolved?

nboldi:
Can we tell if the nondeterminism is in the system or in the tests?

gerazo:
@kovacsgabor aren't the tests get flaky because of using static/singleton stuff which is not destructed/inited properly? It would be a nightmare to put flaky tests on master and see them occasionally failing.

devaigergely81:
@kovacsgabor, do you still find the tests flaky, or did your modifications solve the issue?

devaigergely81:
@nboldi, please synch this branch with master so that we can go on with resolving the flaky test case issue.

nboldi:
@devaigergely81 I synched it with master.

devaigergely81:
Even if I start with a new eclipse, redo the steps of configuring the dev env and use a new workspace, this branch gives me the following errors, both using the mirror or the release target:
~~~~
Description	Resource	Path	Location	Type
Access restriction: The type 'IMarkerToPseudoSelectorMappingProvider' is not API (restriction on required library 'C:\Users\egerdva\Projects\txtUML\workspace_txtuml_dev\.metadata\.plugins\org.eclipse.pde.core\.bundle_pool\plugins\org.eclipse.papyrus.infra.gmfdiag.css_1.1.2.201509161440.jar')	AnimationMarkerToPseudoSelectorMappingProvider.java	/hu.elte.txtuml.diagnostics/src/hu/elte/txtuml/diagnostics/animation	line 6	Java Problem
Access restriction: The type 'IMarkerToPseudoSelectorMappingProvider' is not API (restriction on required library 'C:\Users\egerdva\Projects\txtUML\workspace_txtuml_dev\.metadata\.plugins\org.eclipse.pde.core\.bundle_pool\plugins\org.eclipse.papyrus.infra.gmfdiag.css_1.1.2.201509161440.jar')	AnimationMarkerToPseudoSelectorMappingProvider.java	/hu.elte.txtuml.diagnostics/src/hu/elte/txtuml/diagnostics/animation	line 11	Java Problem
~~~~

@ELTE-Soft/all, what is your experience?

gerazo:
I've tried the full reinstall thing with this on animation branch before merge and there was no such error like this. Such errors are usually solved by putting something from Required-bundle to Import-packages (or vice versa). Were there a papyrus version change between this version and the one where animation was ready?

nboldi:
I thought that only I had that problem from a wrong version of something, because it passed CI tests.

devaigergely81:
The update site generated from this branch cannot be installed. (Please try this with a fresh Eclipse for Java before creating a pull request!)
The error is:
~~~~~
Cannot complete the install because one or more required items could not be found.
  Software being installed: txtUML 0.3.0.201602021408 (hu.elte.txtuml.feature.feature.group 0.3.0.201602021408)
  Missing requirement: txtUML Export 0.3.0.201602021408 (hu.elte.txtuml.export 0.3.0.201602021408) requires 'bundle hu.elte.txtuml.utils.eclipse 0.3.0' but it could not be found
  Missing requirement: txtUML Export 0.3.0.201602021450 (hu.elte.txtuml.export 0.3.0.201602021450) requires 'bundle hu.elte.txtuml.utils.eclipse 0.3.0' but it could not be found
  Cannot satisfy dependency:
    From: txtUML Papyrus Visualizer 0.3.0.201602021408 (hu.elte.txtuml.export.papyrus 0.3.0.201602021408)
    To: bundle hu.elte.txtuml.export 0.3.0
  Cannot satisfy dependency:
    From: txtUML 0.3.0.201602021408 (hu.elte.txtuml.feature.feature.group 0.3.0.201602021408)
    To: hu.elte.txtuml.export.papyrus [0.3.0.201602021408]
~~~~

nboldi:
It must be some kind of merge error. I'll check it out tomorrow.

nboldi:
After the last commit standalone install is working for me

plugins/hu.elte.txtuml.project/src/hu/elte/txtuml/project/wizards/NewTxtUMLModelCreationPage.java
```diff
@@ -139,4 +144,63 @@ protected IStatus packageChanged() {
 		return status;
 	}
 
+	/**
+	 * Checks a package if it belong to an existing model. Searches for a
+	 * package-info.java compilation unit in the package or one of the ancestor
+	 * packages and checks if it has the {@link Model} annotation.
+	 */
+	private static boolean isModelPackage(IPackageFragmentRoot pfRoot, String packageName) throws JavaModelException {
```
devaigergely81: This branch introduces a new warning compared to master, saying that this function is never used. Looks problematic... Please look into it!

plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/ModelClass.java
```diff
@@ -372,28 +367,6 @@ void checkLowerBound(Class<? extends AssociationEnd<?, ?>> assocEnd) {
 
 	}
 
-	/**
```
devaigergely81: This (and all port-related stuff) should be part of the branch.

devaigergely81:
I confirm that the branch's update site can be installed now, but some example projects fail to compile, because port functionality was lost by the merges.

plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/Action.java
```diff
@@ -87,52 +89,6 @@ public static void delete(ModelClass obj) {
 	}
 
 	/**
-	 * Connects two ports through the specified assembly connector.
```
devaigergely81: This is also missing.

plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/Action.java
```diff
@@ -293,20 +253,6 @@ public static void start(ModelClass obj) {
 	}
 
 	/**
-	 * Asynchronously sends the specified signal through the specified reception.
```
devaigergely81: Send to port is missing.

plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/ModelClass.java
```diff
@@ -191,8 +188,6 @@
 	 */
 	private final AssociationsMap associations = AssociationsMap.create();
 
-	private final ClassToInstanceMap<Port<?, ?>> ports = MutableClassToInstanceMap.create();
```
devaigergely81: Missing.

plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/JtxtUMLCompilationParticipant.java
```diff
@@ -62,36 +62,36 @@ public void buildStarting(BuildContext[] files, boolean isBatch) {
...
 	private void validateFile(BuildContext file) {
-		java.io.File systemFile = null;
-		IPath location = file.getFile().getLocation();
-		if (location != null) {
-			systemFile = location.toFile();
-		}
-		if (systemFile == null) {
-			return;
-		}
+			java.io.File systemFile = null;
```
devaigergely81: Look like these changes are introducing tabs. Would be good to get rid of tabs.

plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/visitors/CompositionVisitor.java
```diff
@@ -20,13 +20,13 @@ public CompositionVisitor(TypeDeclaration root, ProblemCollector collector) {
...
 	@Override
 	public boolean visit(TypeDeclaration node) {
-		if (ElementTypeTeller.isContainer(node)) {
-			++containerMembers;
-		} else {
-			++partMembers;
-		}
-		return false;
-	}
+					if (ElementTypeTeller.isContainer(node)) {
```
devaigergely81: Tabs again?

plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/visitors/ModelClassVisitor.java
```diff
@@ -24,7 +24,7 @@
 	public ModelClassVisitor(ProblemCollector collector) {
 		super(collector);
 	}
-
+
```
devaigergely81: Tab?

plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/visitors/Utils.java
```diff
@@ -8,55 +8,52 @@
...
 import hu.elte.txtuml.validation.ProblemCollector;
 import hu.elte.txtuml.validation.problems.general.InvalidModifier;
 import hu.elte.txtuml.validation.problems.general.InvalidTemplate;
 
-/**
- * Utility methods for visitors to check common AST elements.
- */
 public class Utils {
-
+
```
devaigergely81: Tabs?

devaigergely81:
Other than the noted problems above, I liked the changes: This will be a great improvement to our code base once merged back!

nboldi:
Installation is fine, PingPong model is working.

nboldi:
@kovacsgabor Some parts of port implementation were lost & found, and therefore lost their line history. If it acceptable, please merge.

kovacsgabor:
Please, review my updates.

nboldi:
Every change seems OK to me.

kovacsgabor:
I think it would be better if the log on the standard output only contained the message itself and its type (INFO, ERR, WARN, etc.). The date and the logger instance's name are no useful information for the user when reading the standard output in Eclipse. In my opinion, they should only appear in the log file.

nboldi:
OK for me. But I think that the date might be useful.

kovacsgabor:
On the standard output? Why? It is only viewed in Eclipse, not kept for later use.

nboldi:
Standalone execution?

kovacsgabor:
We also print the log into a log file. This is not the one that will be kept in case of a standalone execution? In the file, we have to print the date, that is obvious.

nboldi:
OK, print only the most important information.

kovacsgabor:
Another question: In December we talked about the logging of the modeling API, but I cannot remember the decision (and I could not find it anywhere in our documentations so I fear it was not written).

The problem specifically: In the past, the modeling API differentiated the so-called user log from the log printed by the executor. The first one is to which the user may write by calling the `Action.log` and `Action.logError` methods. In the second one, the executor reports events about the model execution.

In this branch, they are only differentiated by their type (INFO/WARN/ERR).

Should we use two different loggers to separate these logs or it is not required?

I would also consider differentiating the executor's log (which contains information about the model itself) and log (typically error log) printed by other components (which contains information about the txtUML infrastructure and how it runs).

nboldi:
In the meetings we avoided the topic of how logging (both error and simple output) works in the models. We should talk about it, but I think that it should be implemented on the architecture that is designed in this branch.

nboldi:
We can introduce `userInfo`, `userError`, ... methods to `hu.elte.txtuml.utils.Logger` that log into a different `org.apache.log4j.Logger` object if needed.

kovacsgabor:
After discussing with @devaigergely81, I added three logging categories. Please, review these updates.

kovacsgabor:
I tested the whole branch, and I found only one remaining problem: if the user wants to customize the logs, that becomes problematic as log4j only loads the first `log4j.properties` file it founds (it is loaded by the class loader, actually), which seems to be the one in the `utils` package. So the user cannot override our default settings by simply creating their own `log4j.properties` file.

nboldi:
But I think that the location of the log4j file can be overridden by `-Dlog4j.configuration={path to file}`

kovacsgabor:
Yes, that is true. Not a very convenient solution, but works.

devaigergely81:
OK, I agree with the merge after conflicts are resolved.

kovacsgabor:
I am working on it.

Issue #193, opened 2015-12-04
Get rid of Moka


devaigergely81:
Reasons:
- Moka is a Papyrus-based UML execution framework mainly for execution and debugging of activities. We only use a utility class of it in order to animate state machines.
- There is a known issue with Moka: If the model grows huge or references huge models, Moka freezes the application when it searches for all possible diagrams that the given model element might appear on. Since we animate state machine diagrams generated by us, we know exactly which elements are on which diagrams, so we can avoid this performance bug of Moka.
- Later we plan to show the animation of more than one instances of the same class on a single state machine diagram by color coding. In order to enable this, we have to directly control the colors of the animation instead of the Moka defaults.

How:
- Find out the Papyrus API to change the fill/border color or model elements. (Possibly by reviewing the implementation of Moka).
- Change the state machine animation implementation to use that API directly.
- Remove Moka from the dependencies.

gerazo:
I like it

gerazo:
Reminder: As soon as PR203 is in, the wiki needs to be corrected to reflect that Moka installation is no longer needed

gerazo:
And finally shit hit the fan: the proposed solution (getting-rid-of-moka branch) on a freshly installed Eclipse where Moka was never installed. The reason that putting markers on objects is not enough to have animation. Moka probably uses some extension points to inject new functionality into the rendering of objects, but what is sure that it uses CSS stylesheets to parameterize the outlook of the animation. When Moka is not installed, these parts are missing and nothing is seen.
However there are examples how to directly modify the rendering of an object through Papyrus EditorParts.

gerazo:
Issue was fixed in linked PR.

PR #203, opened 2015-12-21
Getting rid of moka

gerazo:
Fix for this: https://github.com/ELTE-Soft/txtUML/issues/193
All dependencies and target settings were removed for Moka. I hope it will make the whole thing a degree lighter.


gerazo:
Please review @devaigergely81, @kmate, @kovacsgabor or anybody who feels empowered.

kmate:
I removed moka from associate sites too with the previous commit. It looks that the diagram animation works in general, but I found two problems:
1. Initial pseudo-states are not highlighted.
2. The markers remain on diagram after the execution is terminated because either the model execution ended or either it was terminated by the user.

I did not tested the update site, I used a guest eclipse.

gerazo:
Thanks @kmate I will check the two items mentioned. I will try to find out what extra functionality is present in the original which can remedy these issues.

gerazo:
Sometimes it is not working at all. It seems like some stylesheets are missing and without moka, nothing happens on sending the markers. I have to thoroughly check on a fresh install but if that is the case, we have to access editor parts directly. :-1:

Issue #173, opened 2015-12-01
Executions in a separate files lead to false errors


devaigergely81:
Commit 2385f9efe72ad3769e042a8ac767056fbd2f5c1e moved the executions outside the model to avoid validation errors.
When these models are freshly imported in a workspace or a clean+build happens, some of the executions are marked erroneous: The types used in them cannot be resolved. It is usually the 'train' or 'machine' example is affected.
If the file is saved after a dummy modification, the errors go away.
The problem is confirmed by @devaigergely81, @kmate, @nboldi, and it also appears if txtUML is installed.
Seems to be an Xtext/Xbase bug.

devaigergely81:
Associated exceptions:
~~~~
16089 [Worker-9] ERROR org.eclipse.xtext.xbase.jvmmodel.JvmTypesBuilder  - Error initializing JvmElement
java.lang.NullPointerException
	at org.eclipse.xtext.xbase.typesystem.references.LightweightTypeReferenceFactory.toPlainTypeReference(LightweightTypeReferenceFactory.java:117)
	at org.eclipse.xtext.xbase.typesystem.references.StandardTypeReferenceOwner.toPlainTypeReference(StandardTypeReferenceOwner.java:108)
	at org.eclipse.xtext.xbase.jvmmodel.JvmTypeReferenceBuilder.typeRef(JvmTypeReferenceBuilder.java:112)
	at hu.elte.txtuml.xtxtuml.jvmmodel.XtxtUMLJvmModelInferrer.calculateApiSuperType(XtxtUMLJvmModelInferrer.java:788)
	at hu.elte.txtuml.xtxtuml.jvmmodel.XtxtUMLJvmModelInferrer.access$7(XtxtUMLJvmModelInferrer.java:659)
	at hu.elte.txtuml.xtxtuml.jvmmodel.XtxtUMLJvmModelInferrer$18.apply(XtxtUMLJvmModelInferrer.java:554)
	at hu.elte.txtuml.xtxtuml.jvmmodel.XtxtUMLJvmModelInferrer$18.apply(XtxtUMLJvmModelInferrer.java:1)
	at org.eclipse.xtext.xbase.jvmmodel.JvmTypesBuilder.initializeSafely(JvmTypesBuilder.java:206)
	at org.eclipse.xtext.xbase.jvmmodel.JvmTypesBuilder.toClass(JvmTypesBuilder.java:384)
[...]

PR #205, opened 2015-12-27
Inferer corrections and optimization

kmate:
This branch was originally created to solve #173, but unfortunately I am unable to reproduce the bug with any other model than xtrain. It could be some kind of a complicated xtext/xbase bug I think...

However I made some minor fixes and optimizations related to several xtxtuml components including the jvm model inferer.


devaigergely81:
I have tested this in an installed setting and reviewed the code changes.

PR #206, opened 2016-01-12
Ports in jtxtuml

kovacsgabor:
@nboldi, @devaigergely81, please review.


devaigergely81:
I have tested this with a new model (soon to be added as an example), it worked OK.
I'll merge.

Issue #210, opened 2016-01-24
Empty interface


devaigergely81:
It would be handy to define an `Empty` interface in the modeling API, to be used as generic argument in one-way port definitions.

kovacsgabor:
Can we name it simply ```Empty```? Wouldn't it be better to name it ```EmptyInterface``` or create it as a nested interface of ```Interface``` to be ```Interface.Empty```? My worry is that we use a too general name which we might use in other cases in the language later.

devaigergely81:
`Empty` is already used for the empty containers, I think. I perfer `Interface.Empty`.

PR #223, opened 2016-01-28
Interface.Empty is added

kovacsgabor:
See #210.
@devaigergely81, please review.


devaigergely81:
Tested in installed setup, works fine.

Issue #109, opened 2015-11-04
Author tags?


kovacsgabor:
Do we wish to keep the author tags in JavaDoc? Or to add new in the future? Their use is not consequent at all.

nboldi:
I see no use for them. I think they only encourage personal code-ownership instead of shared ownership.

kovacsgabor:
@ELTE-Soft/all, what is your opinion?

kmate:
I agree with nboldi, I never use them. Also, Git already tracks all changes with author information on a more fine-grained level...

gerazo:
I think the only name(s) which makes sense in a javadoc are the developer(s) to whom I can turn to if I have technical questions. It is more like knowledge ownership instead of the classical one for which we are all responsible. The closest term which comes to my mind is "maintainer(s)".

dobreffandras:
Although I can fully agree with @gerazo  to keep the reliability and consistency we strictly need to stick to some rules. I doubt it can be done with 100% success, and if it's not used precisely can easily cause confusion. This case is the typical "Not doing at all is better than doing it wrong" in my opinion.

nboldi:
It would be better if we would just assign each plugin project to one of the members of the team, instead of littering all the classes with author tags. For example, the model.api plugin should be overviewed by @kovacsgabor, the releng projects by @gerazo, the validator by myself.

devaigergely81:
I think the conclusion is to remove the author tags. Someone who is bored, please take this issue. :-)

PR #224, opened 2016-01-28
Remove author tags

kovacsgabor:
See #109.
Created from ```fix/issue212``` branch because that is the only small branch which has an open pull request and added new author tags.


devaigergely81:
This also includes #220, so needs real testing.

devaigergely81:
Works fine.

PR #225, opened 2016-01-28
Use dotted qualified names in assoc. nav. expressions

### LINKED ISSUE #204: Navigation expressions should use dotted qualified names
Currently the navigation expression specifies association names with double colons (::). They should be specified by ordinary qualified names, which are separated by dots (.). This change enables simplifications in qualified name handling.

However, the implementation of this feature is not quiet straightforward, as the navigation expression should be converted into an abstract feature call in the grammar. The following snippet shows an example how to do this:

~~~~
XMemberFeatureCall returns xbase::XExpression :
	XPrimaryExpression
	(
...
	| 
		=>({RAlfAssocNavExpression.target=current} '->') feature=[types::JvmIdentifiableElement|ValidID]
	)*
;
~~~~

This solution results in a conflict-free grammar, but the type computations needed are not trivial, as it also involves scoping and linking services. I consider this as a relatively hard task, so we should decide whether to implement this in the future.

djnemeth:
This is intended to fix #204. For a brief overview, see the (full) commit messages.
@devaigergely81 and @kmate, please review.


kmate:
These changes seem reasonable, however, I did not try to compile or run this branch.

One little thing: in the type computer, instead of
~~~
if (rightChild == null || rightChild.name == null)
~~~
I think you can either write
~~~
if (rightChild?.name == null)
~~~
in Xtend syntax (null-safe access). But the current expression is maybe more clear.

djnemeth:
As we already use the `?.` operator in other parts of the implementation, I agree that the simplified syntax is reasonable here as well. Thank you!

devaigergely81:
Works fine.

PR #236, opened 2016-02-05
Fix "unused import" warnings

### LINKED ISSUE #226: Incomplete type usage collection in XtxtUML
Although the linking itself occurs on a much higher abstraction level, Xtext provides a [way](https://eclipse.org/Xtext/documentation/301_grammarlanguage.html#syntax) to define cross-references to language elements within the grammar. For example, consider the following parser rule from the XtxtUML grammar: 

```
TUTransitionTrigger :
	'trigger' trigger=[TUSignal|QualifiedName] ';'
;
```

Here, the `trigger=[TUSignal|QualifiedName]` part is an example of the aforementioned grammar-level cross-reference syntax, indicating that the trigger of a transition is an EClass instance of type TUSignal, which EMF type is inferred from the parser rule having the same name.

We use this feature throughout the XtxtUML grammar, as it provides out-of-the-box type validation and content assist support. It also potentially makes some aspects of the runtime faster, as using the `JvmTypeReference` parser rule as an alternative would introduce an additional wrapping layer to be built and maintained for each language-level cross-reference.

<hr>

In the current implementation, however, in an XtxtUML source file, if an _imported_ type is only referenced in abovementioned, grammar-level cross-references, a false warning is issued on the import stating that the type is never referenced.

One can easily reproduce the warning, for example in the following way. Given an XtxtUML model `M` with a class `A`, create a new .xtxtuml file with the following contents:

```
package m;

import m.A;

class B extends A;
```

Here, a warning is shown on the import with the message _"The import 'm.A' is never used."_. Naturally, if the import is removed, the `m.A` class can't be referenced using its simple name, causing an error marker to appear with the message _"Couldn't resolve reference to TUClass 'A'."_.

<hr>

This is caused by the fact that `TypeUsageCollector` – the class responsible for collecting the used types in an Xbase resource – ignores such references, as it can be seen in its [`collectAllReferences`](https://github.com/eclipse/xtext/blob/003a5602d3ddb86fd2204e4e6276e5b109ccdd89/plugins/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/imports/TypeUsageCollector.java#L175) method. I can think of 3 possible solutions:

1. Overriding the aforementioned `collectAllReferences` method to include grammar-level cross-references. I consider this solution quite inelegant, as it would involve direct code copying of the original method (assuming we wouldn't want to iterate through the contents of the resource once again) and other, private auxiliary functions defined in the original class.

2. Intervention in the type usage collection in some other way, for example by extending (but not modifying) the functionality of the overridden `collectAllReferences` method by iterating through the contents of the given resource again. This is a cleaner, but definitely less efficient solution than the first one.

3. Replacing the grammar-level cross-references with `JvmTypeReference`s in the grammar. In this way we would lose the out-of-the-box type validation and content assist support, making the manual implementation of these features necessary. The type computation of some expressions would have to be modified as well. The involved expressions are the following: [`TUClass`](https://github.com/ELTE-Soft/txtUML/blob/e3a8663663e5af26d0430de49bfec5ad943fc130/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/XtxtUML.xtext#L39), [`TUTransitionTrigger`](https://github.com/ELTE-Soft/txtUML/blob/e3a8663663e5af26d0430de49bfec5ad943fc130/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/XtxtUML.xtext#L114), [`TUTransitionVertex`](https://github.com/ELTE-Soft/txtUML/blob/e3a8663663e5af26d0430de49bfec5ad943fc130/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/XtxtUML.xtext#L118), [`TUAssociationEnd`](https://github.com/ELTE-Soft/txtUML/blob/e3a8663663e5af26d0430de49bfec5ad943fc130/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/XtxtUML.xtext#L132), [`RAlfAssocNavExpression`](https://github.com/ELTE-Soft/txtUML/blob/e3a8663663e5af26d0430de49bfec5ad943fc130/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/XtxtUML.xtext#L469).

<hr>

@devaigergely81 and @kmate, what do you think of these solutions? @kmate, can you think of any other possibilities?

djnemeth:
For a brief overview, see the (full) commit messages. For additional details, see #226. This pull request also resolves #226. @devaigergely81 and @kmate, please review.

### DIFF
--- a/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/XtxtUMLRuntimeModule.java
+++ b/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/XtxtUMLRuntimeModule.java
@@ -6,6 +6,7 @@ import org.eclipse.xtext.scoping.IScopeProvider;
 import org.eclipse.xtext.scoping.impl.AbstractDeclarativeScopeProvider;
 import org.eclipse.xtext.validation.ConfigurableIssueCodesProvider;
 import org.eclipse.xtext.xbase.compiler.XbaseCompiler;
+import org.eclipse.xtext.xbase.imports.ImportedTypesCollector;
 import org.eclipse.xtext.xbase.scoping.batch.ImplicitlyImportedFeatures;
 import org.eclipse.xtext.xbase.typesystem.computation.ITypeComputer;
 import org.eclipse.xtext.xbase.typesystem.internal.DefaultReentrantTypeResolver;
@@ -17,6 +18,7 @@ import com.google.inject.name.Names;
 
 import hu.elte.txtuml.xtxtuml.compiler.XtxtUMLCompiler;
 import hu.elte.txtuml.xtxtuml.compiler.XtxtUMLGenerator;
+import hu.elte.txtuml.xtxtuml.imports.XtxtUMLImportedTypesCollector;
 import hu.elte.txtuml.xtxtuml.jvmmodel.XtxtUMLTypesBuilder;
 import hu.elte.txtuml.xtxtuml.naming.IPackageNameCalculator;
 import hu.elte.txtuml.xtxtuml.naming.XtxtUMLPackageNameCalculator;
@@ -85,4 +87,9 @@ public class XtxtUMLRuntimeModule extends AbstractXtxtUMLRuntimeModule {
 	public Class<? extends ConfigurableIssueCodesProvider> bindConfigurableIssueCodesProvider() {
 		return XtxtUMLConfigurableIssueCodes.class;
 	}
+
+	public Class<? extends ImportedTypesCollector> bindImportedTypesCollector() {
+		return XtxtUMLImportedTypesCollector.class;
+	}
+
 }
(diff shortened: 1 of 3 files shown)


kmate:
It is a very unfortunate thing that `XtxtUMLImportedTypesCollector` must be kept in sync with the grammar manually. I hope there will be a better solution later... However, it seems ok. I have not tried it.

gerazo:
When will be the plugin.xml warnings fixed? Are they fixable at all?

djnemeth:
Jenkins, retest this please.

djnemeth:
The plugin.xml of the ui module is based on the one generated by Xtext. I'm afraid as long as the framework uses deprecated extension points, the warnings will stay. I'll check if they can be actualized.

devaigergely81:
Tested, works fine.

PR #238, opened 2016-02-09
CPP overlong merge

gerazo:
We believe that the very minimal criteria is reached. It was tested on fresh Eclipse install as well.


gerazo:
@devaigergely81 What do you think?

devaigergely81:
Cool!

PR #240, opened 2016-02-11
Model export fix

nboldi:
Action code export into activity diagrams


nboldi:
@devaigergely81, @kovacsgabor: Please Review!

nboldi:
Restructuring of models created a problem for export. Working on it.

nboldi:
@devaigergely81, turned activity generation off when exporting diagrams. Check if ready to merge.

kovacsgabor:
jenkins, retest this please

nboldi:
Testing problems seem to be fixed. I ran clear-verify for a few times and worked all time.

devaigergely81:
Machine example:
- The `Machine.doWork()` method's behavior is exported as opaque behavior instead of activity.
- String parameters of the log instructions are present in the model, but the `log` call is missing.

devaigergely81:
The activity generated from the navigation expression 
```Machine m = User.this.assoc(Usage.usedMachine.class).selectAny();```
seems problematic:
navigatebug
The read self action should be the first one, and its result is not propagated.

**Comment by devaigergely81** [comment]
In case of
```
Action.send(User.this,new DoYourWork());
```
the request inside the send object action seems duplicated:
sendbug

**Comment by nboldi** [comment]
> The activity generated from the navigation expression seems problematic:
The result is passed to the link end data by the first object flow. The order is still wrong, I fix it.

**Comment by nboldi** [comment]
> The User.doWork() method's behavior is exported as opaque behavior instead of activity.

It is exported both as opaque behavior and activity. I did not stop generating opaque behaviors in case they are needed for historical reasons, but both the activity and the opaque behavior is registered as a method of the operation.

**Comment by nboldi** [comment]
Unfortunately, there are still problems with JDTs classpath. I produced a situation where `ModelClass` is resolved for `Machine` but not resolved for `User`, two classes in the same package :angry:

**Comment by nboldi** [comment]
Interesting, the classpath seems to be correct, but `resolveBinding()` returns `null` for `User`.

**Comment by nboldi** [comment]
Oh, it seems there is an error in the Machine example that killed both validation and export:
```
 The type hu.elte.txtuml.api.model.assocends.ContainmentKind$SimpleEnd cannot be resolved. It is indirectly referenced from required .class files
```

**Comment by nboldi** [comment]
Nevermind. Workspace got corrupted somehow.

**Comment by nboldi** [comment]
> In case of `Action.send(User.this,new DoYourWork());` the request inside the send object action seems duplicated:

I could not find the doubled representation. The request of a `SendObjectAction` does not even have multiplicity. It may be a visualization error.
```xml
        <executableNode xmi:type="uml:SendObjectAction" xmi:id="_Xzj7kNyBEeWOluVIInnbnQ" name="send_ButtonPress_to_myMachine">
          <request xmi:type="uml:ValuePin" xmi:id="_aXTDoNyBEeWOluVIInnbnQ" name="send_ButtonPress_to_myMachine_request" incoming="_aXTDodyBEeWOluVIInnbnQ" type="_WBlnpNyBEeWOluVIInnbnQ"/>
          <target xmi:id="_fSaSwNyBEeWOluVIInnbnQ" name="send_ButtonPress_to_myMachine_target" incoming="_fSaSwdyBEeWOluVIInnbnQ" type="_WBlnodyBEeWOluVIInnbnQ"/>
        </executableNode>
```

devaigergely81:
Thanks for the fixes and clarifications so far! Those issues seem to be OK now.

devaigergely81:
In case of
```
Machine m = Action.create(Machine.class);
Action.delete(m);
```
I cannot see in the generated model how is the variable `m` formally connected to the target of the delete object action. The target has the same name as the variable, but there is no dataflow edge and, in the XML, the ID of the target is never mentioned elsewhere.

nboldi:
Fixed the missing link in object deletion.

devaigergely81:
The delete object action seems OK now.

devaigergely81:
In case of diagram generation, no actions should be generated. Operation bodies are OK, they are not generated (only as opaque behaviour), but state machine entries, exits and effects are generated.

nagyandras95:
What is missing for the branch to merged back into master? I need it to start activity generation.

nboldi:
> In case of diagram generation, no actions should be generated. Operation bodies are OK, they are not generated (only as opaque behaviour), but state machine entries, exits and effects are generated.

The generation of entry and effect actions on states, and effects and guards on transitions are turned off when generating diagrams.

nboldi:
I know about a few problems, but they are more easily corrected along a big refactoring, so I think the branch should be merged soon.

devaigergely81:
All problems I found are now resolved, thx!

Issue #241, opened 2016-02-12
Missing multiplicity handled incorrectly


devaigergely81:
If association end multiplicity is omitted in XtxtUML, there is no error shown in the editor, but invalid JtxtUML code is generated and model export fails with an exception.
Two things need to be fixed:
- XtxtUML: Either make the multiplicity mandatory in the grammar or fix the Java code generation such that it generates `One` as default multiplicity.
- Model export should provide meaningful error message in case of problems in the model.

djnemeth:
What a nasty bug indeed. I'm fixing the XtxtUML part ASAP.

djnemeth:
What is the status of the model export part?

kovacsgabor:
@nboldi, what can you say about the model export part of this issue? As I see, the X part is done.

djnemeth:
Currently, the model exporter throws an exception and shows the problematic code snippet and the stack trace if an invalid (or not yet supported) model element is encountered, for example:

![model-exporter-error](https://cloud.githubusercontent.com/assets/13290204/17360352/731fd50e-596b-11e6-9a83-785aaf2559e5.png)

@devaigergely81, do you consider this enough to close this issue?

djnemeth:
The team have decided that in the current state of the implementation, these kind of error messages are sufficient.

PR #243, opened 2016-02-13
Fix handling of omitted association end multiplicity in XtxtUML

djnemeth:
As it has been reported in #241, if an association end multiplicity was omitted in XtxtUML, the incorrect type inference lead to faulty JtxtUML code generation. This is now fixed by inferring `One` as default multiplicity.


djnemeth:
@devaigergely81, please review.

devaigergely81:
Works fine.

Issue #242, opened 2016-02-12
Association line touches independent class


devaigergely81:
![diagrambug](https://cloud.githubusercontent.com/assets/7555116/13013715/dacb8344-d1b0-11e5-83f6-d7c901b75224.png)

@GregoBalu, @dobreffandras, please look into this!

GregoBalu:
So I checked my part and the grid results are OK in a way that we would expect it. Somewhere in the calculation to pixel values should be some rounding errors I believe.

I haven't checked the papyrus part, so I'm not sure where does it go wrong.

Still I have thought of this for a while: if I (layout.visualizer) should return exact pixel values for everything. Currently the width/height of the boxes are altered by me anyway. Maybe I should be in charge of these so any placement issues should be considered my fault.

GregoBalu:
This also leads us to better handling lines during the link arrange section. As I mentioned before we need methods to prefer certain routes over others. In this case the link should be in the middle of the corridor between the boxes using 2 turns in my opinion.

PR #246, opened 2016-02-15
Pixel and Grid Widht/Height of boxes adjusted.

GregoBalu:
Fixes issue #242 .


devaigergely81:
I have tested this fix with the following results:
- The diagram fragment posted in #242 looks like this with this branch:
diagrambug2

Problem seems to be solved, but I think the line is closer to the box than a grid unit.
- Now, by adding 1 character to the name of the top left box, we get:
diagrambug3

The line is touching the box again.

See the test project for experimentation: 
[Test.zip](https://github.com/ELTE-Soft/txtUML/files/131495/Test.zip)

GregoBalu:
OK, so the fix I did was indeed needed, but one more thing used the grid/pixel ratio. The papyrus component when scaling, positioning and transforming the diagram pre-drawing. There the same error existed: basically the pixel/grid value was smaller than the actual value, therefore the diagram generated was squashed a bit.
After I fixed this, the given Test model seems to work fine whether you insert any number of 'a' letters into ones name or not.

devaigergely81:
The fix seems to be OK.

devaigergely81:
The branch is ready for merge.

Issue #247, opened 2016-02-15
New repo structure


gerazo:
Create new folder structure:
dev/ plugins releng tests features
examples/ demo tests

And adjust Jenkins to accommodate the new situation

gerazo:
We still need to
- remove .settings
- move exmaples to demo folder

kovacsgabor:
Do we need `.gitignore` files in every project? Woudln't it be simpler if we used the `.gitignore` file in the root folder to ignore every `bin`, `src-gen`, `gen` etc. folders globally?

gerazo:
Well, we could do that. However, I wouldn't like to accidentally miss the creation of a gen folder where it is not supposed to be. The individual files are custom per project and only point to a certain place in a folder. They do not generally ignore every gen folder. I think it is better this way.

gerazo:
demo examples should be: clock, machine, monitoring, pingpong, producer_consumer, train
The decision was to commit .settings still

kovacsgabor:
What about the projects in which the `.settings` folder is already ignored?

gerazo:
As far as I understand, the decision was that the project maintainer can decide to not commit the .settings if she/he sees a special reason for that. I consider that this is the case at these projects.

gerazo:
https://github.com/ELTE-Soft/txtUML/pull/251 solved the remaining problem.

PR #248, opened 2016-02-16
Repo layout change

gerazo:
The tests will fail until Jenkins is not set up correctly. I will only do this if the review is successful.
Please review @devaigergely81 @nboldi @kovacsgabor. If you like it DO NOT merge! I need to set Jenkins first.

Every example was moved to tests. Which should we put into demo? (Machine, Train, Monitoring, Producer Consumer only?)
https://github.com/ELTE-Soft/txtUML/issues/247


nboldi:
tests should not go into the plugins directory, but remain in a tests directory under dev

gerazo:
Sorry, I was confusing. Every example was put into examples/tests and some examples are waiting to be promoted into examples/demo because their quality is outstanding.

nboldi:
.settings folders should not be versioned

nboldi:
This renaming does not seem to be right:
tests/hu.elte.txtuml.utils.tests/.classpath → .../hu.elte.txtuml.api.deployment/.classpath

gerazo:
.settings folders were always versioned before
The 2 files mentioned are identical so their hash is the same. Git thinks a different but equally good renaming combination compared to what I did. It is not a problem. Check out the repo and see if everything is still at the same place.

nboldi:
It is strange that git still tries to figure out the renamings even that you manually renamed the files...
The history of identical files could be swapped by the different renaming combination.

gerazo:
Yes, despite the fact that I used "git mv" to move. I think the only way to force distinction is to make all moves a separate commit... which I obviously don't want to do :pig:

kovacsgabor:
https://git.wiki.kernel.org/index.php/GitFaq#Why_does_Git_not_.22track.22_renames.3F

> Git has a rename command git mv, but that is just for convenience. The effect is indistinguishable from removing the file and adding another with different name and the same content.

gerazo:
So, can I consider it reviewed then? :question:

nboldi:
OK for me. :+1:  But the `.settings` folders should not be versioned, so please remove them after the change. They contain eclipse view settings for example. I already added gitignores for them in several plugins.

gerazo:
@nboldi Should I do it in a separate pr?

gerazo:
And can I put Machine, Train, Monitoring, Producer Consumer into the demo folder?

nboldi:
Yes, use a separate branch, or do it on the master.

gerazo:
Ok, we can put the examples to the right place later as well. I start setting Jenkins up. Thanks :cocktail:

gerazo:
retest this please

PR #249, opened 2016-02-16
fix for the reflexive associations problem

devaigergely81:
This fixes the problem of misplaced reflexive association lines. The problem was the following: The original implementation of the `setConnectionBendpoints` function tried to restore the start and endpoint of a line from the anchor points. For some reason this works fine for non-reflexive lines, but gives (0,0) points for reflexive ones, which triggers Papyrus to move (and to distort) the whole line.

@dobreffandras, please review this branch! If you think its OK, please merge it to master and from there to the layout-sm-diagrams branch to see if the distorted lines of state machines could be resolved similarly.

### DIFF
--- a/dev/plugins/hu.elte.txtuml.export.papyrus/src/hu/elte/txtuml/export/papyrus/api/DiagramElementsModifier.java
+++ b/dev/plugins/hu.elte.txtuml.export.papyrus/src/hu/elte/txtuml/export/papyrus/api/DiagramElementsModifier.java
@@ -100,24 +100,22 @@ public class DiagramElementsModifier {
 	}
 	
 	/**
-	 * Sets the BendPoints of a connection. (Starting point and Ending points are not BendPoints)
+	 * Sets the points of a connection.
 	 * @param connection - The connection
-	 * @param bendpoints - The BendPoints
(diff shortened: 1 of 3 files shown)


dobreffandras:
It's a very interesting problem and solution, I did not thought about it, and still not fully convinced why does it work. 
Anyway the two branches ('fix-diagram-lines' and 'layout-sm-diagrams') has many differences and I did not have the strength for a merge yet. Instead I applied the same modifications on the file but the problem wasn't solved on 'layout-sm-diagrams'.

Conclusion: I have to take further investigation of this solution on this branch. After that it can be merged back to the 'master'. But afterwards merging the modifications from 'master' back to 'layout-sm-diagrams' has to be done very carefully. 
Since the problem couldn't be solved on the 'layout-sm-diagrams' branch, I have the feeling that the problem may have different sources.

devaigergely81:
These changes alone will not solve the problem. Fix for the scaling issues in #246 are also necessary. My experiments with the 'layout-sm-diagrams' branch today show that there are scaling issues on that branch regarding the `y` coordinates of link points.

devaigergely81:
This is already part of another branch, so I close this.

PR #261, opened 2016-02-25
Fixed wrong use of the composite aggregation kind UML term.

### LINKED ISSUE #259: Aggregation kind: composite
The UML term `composite` (as an aggregation kind) is used to mark the non-container end of a composition which is the exact opposite as it is currently used in the modeling API.

kovacsgabor:
To fix #259.


devaigergely81:
Works fine, code changes seem OK.

PR #262, opened 2016-02-25
Changed order of port parameters

### LINKED ISSUE #222: Change the order of the parameters of a port?
In the current JtxtUML implementation, a port has two parameters, its provided and required interface, in this order. However, it came up, that maybe the opposite order is in some way more natural to a beginner user and therefore it is a question whether to change this order or not.

kovacsgabor:
To fix #222.


devaigergely81:
After my fix the pingpong and clock examples run OK. I think this is now OK to merge. @kovacsgabor, please review my change and merge if you think it OK.

kovacsgabor:
The fix is OK. Sorry for this mistake.

PR #263, opened 2016-02-25
Delegation connectors

### LINKED ISSUE #215: Differentiate between assembly and delegation connectors in JtxtUML?
It was previously discussed on our meetings that maybe the two kinds of connectors (assembly and delegation) need a different name (different base class) in JtxtUML.

kovacsgabor:
To fix #215.


kovacsgabor:
Contains #261.

devaigergely81:
OK.

PR #264, opened 2016-02-25
InPort and OutPort

### LINKED ISSUE #239: InPort and OutPort in JtxtUML
Based on @devaigergely81's proposal, the following two specific port types could be introduced in JtxtUML:

a)
`InPort<P>`, which would be equivalent to a `Port` which's required interface is `P` and provided interface is `Empty`.
b)
`OutPort<P>`, which would be equivalent to a `Port` which's required interface is `Empty` and provided interface is `P`.

They could still be behavior ports without the introduction of any new types as behavior ports are marked with an annotation.

kovacsgabor:
To fix #239. Contains #262.


devaigergely81:
When trying to run the clock example in this branch, I keep getting an error like this:
```
[EXEC ERROR] Error: creating a model object of the type Clock has failed with the following parameters: 17 (Integer), 20 (Integer), 3 (Integer)
```

kovacsgabor:
I fixed the reported problem.

devaigergely81:
Works now.

PR #270, opened 2016-02-29
Swapped parameters of Action.send

### LINKED ISSUE #268: Changing the parameter order of Action.send
The reason of this proposal is to bring the JtxtUML and XtxtUML closer, as in XtxtUML we write a send operation like this: `send new S() to obj` while in JtxtUML `Action.send(obj, new S())` would be called.

kovacsgabor:
See #268.


devaigergely81:
Works OK.

PR #284, opened 2016-03-08
Externals in JtxtUML

### LINKED ISSUE #70: External classes: attributes (fields) or associations?
We have external classes in JtxtUML models but we cannot keep references to them as both possible ways to do this are currently forbidden in JtxtUML:
 1. To connect them to model classes through associations. To be honest, it is not "forbidden", but rather "unsupported". When associations were implemented, we did not even think about this case.
 2. Creating attributes (fields) of model classes that could save references of ExternalClass types. This is forbidden by validation.

So the question is: Should external classes become classes of the model with "full rights", that is, they could be part of associations or should they look more like data types by keeping references to them in attributes (fields)?

If we prefer the first case, another problem will come up: multiplicities. What multiplicities could an association end have which is connected to an external class? For example, is it allowed to have a multiplicity of 1..1? Or 1..*? If yes, this leads to more runtime validations needed.

### LINKED ISSUE #71: External classes: Java classes or interfaces?
The question is, whether external classes of a JtxtUML model should be implemented as Java classes or interfaces extending or implementing the ```ExternalClass``` interface.

Currently, they are classes which do not have to be part of the model "physically", that is, they do not have to be nested classes of the main class of the model (the one that extends ```Model```).

I think it would be better if they became full parts of the model. So after #56, they would have their own files inside the package structure of the model.

However, in this case, it could make things more clear if they were interfaces as their implementation is not really part of the model. Instantiating them could still remain possible through static methods of these interfaces. For example:

```java
// inside the model
interface Ext extends ExternalClass {
  static Ext create() {
    return new ExtImpl();
  }
  void a();
  void b();
  // ...
}

// outside the model
class ExtImpl implements Ext {
  void a() { /* ... */ }
  void b() { /* ... */ }
  // ...
}
```

### LINKED ISSUE #72: Standard library as one or more models?
If in #71 we choose to make external classes parts of the model, it will become a problem that external classes (currently only timers) defined in the ```hu.elte.txtuml.api.stdlib``` project are not part of the user's model and therefore not allowed to be used there.

I suggest that we make these external classes part of a txtUML model (or more models in the future) which then could be imported to the user's model.

### LINKED ISSUE #254: External class support in JtxtUML
JtxtUML needs a full support for external classes (and later, for external data types).

Related issues: #70, #71.

kovacsgabor:
The implementation is based on the solution proposed by @devaigergely81 in #254.

This pull request should only be merged if the questions about this proposal (see at the issue) are answered as based on the answers this solution may need some changes.

Details:
 * Removes the old `ExternalType` and `ExternalClass` interfaces.
 * Introduces three annotations:
  * `@External`: annotated elements are not part of the model.
  * `@ExternalBody`: body of the annotated method or constructor is not part of the model.
  * `@ExternalSuperInterface`: a repeatable annotation with an array of Java classes to mark that some super interfaces of a class are not part of the model.
 * Updates the validation and the examples according to this change.
 * Resolves #254.
 * Resolves #70.
 * Resolves #71.
 * Resolves #72.


kovacsgabor:
jenkins, retest this please

devaigergely81:
It seems that this is still under development, with failing tests, so I close the pull request for now.

Issue #260, opened 2016-02-25
Refactor uml export


nboldi:
The codebase is inconsistent and I think is much more complex than it should be. I would try to make the code more clear to discover hidden bugs and edge cases.

PR #302, opened 2016-03-11
Ports uml export

nboldi:
Export of ports to UML is implemented, with a few problems:
 - Somehow every connector turns to be assembly connector. I did not dug into the logic of this calculation. The API says: `a connector with one or more ends connected to a Port which is not on a Part and which is not a behavior port is a delegation; otherwise it is an assembly.` But I cannot understand how will this be realized.
 - The connect action is flawed. It only receives the ports as values. But changing this would cause a mess, so I would like to stick with it until I can rework the architecture (aka #260)


nboldi:
Because pingpong cannot be exported successfully, I tested the project with a modified version, where action code not related to ports is removed.

nboldi:
Will be reimplemented in the refactored version

PR #309, opened 2016-03-23
Uml export hotfixes

nboldi:
Three simple fixes for three uml export issues.


nagyandras95:
Is the request of `SendObjectAction` necessary to be a value pin? It does not contain `ValueSpecification` but it gets its value from an `OutputPin` through an `ObjectFlow`. Or could it contain a `ValueSpecification` like a literal integer not just an object?

nboldi:
I have no control over the created input pin type. A literal expression is exported as a [ValueSpecificationAction](http://download.eclipse.org/modeling/mdt/uml2/javadoc/3.1.0/org/eclipse/uml2/uml/ValueSpecificationAction.html), that has a result output pin.

devaigergely81:
The code changes seem OK to me.
@nagyandras95, please review if this works as expected and do merge if yes.
Thx.

nagyandras95:
It seems OK to me too.

PR #310, opened 2016-03-30
Cpp tests

gerazo:
CPP Export is tested on 4 selected examples for normal cpp source generation. Environment is detected for build system capabilities and the source is built with gcc and clang in debug and release builds (4 examples x 2 compilers x 2 builds = 16 CPP builds).


gerazo:
retest this please

PR #311, opened 2016-04-01
Cpp tests

gerazo:
CPP Tests do a simple cpp generation and also generation with the runtime.
As an extra, they detect the environment and also try cpp compilation with all accessible compilers. cmake and ninja is needed, gcc and clang is detected.
These tests are very flaky in nature, so extra hardening is used. Now they run in maven and also from Eclipse unit testing.


gerazo:
retest this please

PR #314, opened 2016-04-03
Fixes in the JVM model inferrer

### LINKED ISSUE #173: Executions in a separate files lead to false errors

### LINKED ISSUE #313: Incorrect JVM model inference of ports and association ends
Xtext's internal JVM types Ecore metamodel is used to represent structural concepts of the JVM. In XtxtUML, the class [`XtxtUMLJvmModelInferrer`](https://github.com/ELTE-Soft/txtUML/blob/master/dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/jvmmodel/XtxtUMLJvmModelInferrer.xtend) is responsible for mapping XtxtUML entities to the aforementioned metamodel, using the provided JVM model inferrer API.

The inference process has two phases:

1. Creating the skeletons of referable, class-like entities.
2. Inferring the members of these skeletons.

This decoupling can be leveraged to simulate Java's loose declaration order policy (i.e. given a referable language element and a cross-reference to it, the element's definition doesn't have to precede the cross-reference). Indeed, in the first phase we create the skeletons of referable language elements (classes, signals, states and so on), and in the second phase we populate these skeletons, being ensured that all involved cross-references will be valid, as the referable skeletons have already been constructed in the previous phase.

In the current implementation, however, ports and association ends are *not* inferred in the first phase. As a result, if a cross-reference to a port or association end (e.g. in a connector) precedes the actual definition, the generated Java files will be malformed.

To reproduce the issue, build the following Eclipse project: [incorrect-inference.zip](https://github.com/ELTE-Soft/txtUML/files/200984/incorrect-inference.zip).

### LINKED ISSUE #336: If a port is defined after it is referenced in a transition in X, the generated J code is erronous
It should be either fixed or validated.

djnemeth:
This is intended to fix #313 and resolve #173. Also fixes #336.
For a brief overview, see the mentioned issues and the (full) commit messages.


djnemeth:
@devaigergely81 and @kmate, please review.

kmate:
Works for me.

djnemeth:
I suggest not to merge this pull request for now.

I think I have found a solution for #173, which requires the model inferrer to be modified. As such, it could be conveniently implemented within this pull request.

djnemeth:
3936c3f hopefully fixes #173.
@kmate and @devaigergely81, could you please re-review the changes?

kmate:
It works with the test project attached to #173. However, I have not tested it with other configurations.

kovacsgabor:
@djnemeth, please resolve the conflicts of this branch!

kovacsgabor:
Installed, tested, the code seems to make sense.

I think this is ready to be merged.

kovacsgabor:
Although I do not like the fact that `Tester`s can only be run from the src-gen folder. But the execution syntax will have to be changed anyway after #331 so I suggest to merge this for now.

kovacsgabor:
It also fixes #336.

PR #319, opened 2016-04-11
Validation improvements in XtxtUML

djnemeth:
**This pull request:**
- modifies the structure of the XtxtUML validator — by using inheritance (see the first commit message to details) — to separate validation levels,
- introduces new validations rules and
- refactors several existing ones.

**New validations:**
- `sigdata` expression can be used only if the enclosing state or transition is not directly (i.e. through pseudostates «initial and choice») reachable from an initial state,
- only the owner class of a port can send signals to it (i.e. if the target of a send expression is a port, the send expression itself must be inside the owner class),
- declared and expected package names must match,
- constructors must be unique (considering their parameter list),
- only one initial state per statemachine level is allowed,
- at most one entry and exit activity must be defined in a state,
- exactly one initial transition is required for each initial state,
- transitions members (`from`, `to`, etc.) cannot be defined multiple times,
- port members (`required` and `provided` interface) cannot be defined multiple times,
- a constructor's name must match the name of the enclosing class,
- if other states are present, an initial state should be defined (warning),
- for each pseudostate at least one outgoing transition must be defined,
- states should be reachable from the initial state (warning),
- states and transitions must be defined only in classes and composite states,
- pseudostates must not define entry and exit activities,
- `from` and `to` are mandatory for all transitions,
- `trigger` is mandatory for all transitions whose source is not a pseudostate,
- outgoing transitions from pseudostates must not have `trigger` and `port` members,
- transitions must not go in initial states,
- the outgoing transition from an initial state must not have a guard,
- the source and target vertex of a transition must be defined in the transition's enclosing element (e.g. transitions must not cross statemachine levels).


djnemeth:
@kmate, please review. @kovacsgabor, please verify the consistency of the new validation rules with our modeling concepts.

kmate:
Seems ok.

Maybe one thing I want to note: do not collect the utility methods always at the end of a class, but directly after their usage. For example, in `XtxtUMLClassValidator` the method `missingInitialState` is defined on line 244, but used on line 66 and 73. It is not a problem when somebody reads the code in an IDE, and able to jump to the definition and back, but in any other cases it is easier to read the code when the utility method is right after its client methods. Do not change it now, just consider this for future developments.

(Oh, one more thing: if it is easy to write unit tests for validation, we should do it - maybe they could be written in a self-documenting BDD-style. Then we would always have a live list of implemented validations directly in the code base.)

djnemeth:
(I'm planning to implement a comprehensive unit test suite for all XtxtUML components in the upcoming weeks as part of the finalization process of my BSc thesis work. Thanks for the tip about placing utility methods, I often think about this question. In this case I chose to separate validations methods, hoping that in this way it will be more straightforward to see available validations. Nevertheless, I'll take your advice.)

kovacsgabor:
Note:
> * only the owner class of a port can send signals to it (i.e. if the target of a send expression is a port, the send expression itself must be inside the owner class),

This rule could be even stricter: only the the owner _object_ may send signal to the port.

kovacsgabor:
I installed txtUML from this branch in a fresh Eclipse and tested all the above listed validations. They all work and are consistent with our modeling concepts.

I think this is ready to be merged.

PR #323, opened 2016-04-12
Revert "Removed parameter of getSignal methods (also in XtxtUML generation and examples)."

### LINKED ISSUE #322: Revert the removal of getSignal parameter
In #139, we decided to remove the parameter of `getSignal` because it seemed unnecessary with the JDT translation. However, we agreed to forbid casting in (J)txtUML models and therefore the use of `getSignal` without a parameter is problematic at least because the user would be confused where to use and where not to use casting. Despite the fact, that
```java
((MySignal) getSignal())
```
in the current syntax has exactly the same result as
```java
getSignal(MySignal.class)
```
in the old and future syntax because `getSignal` is generic and it casts its result to its type parameter.

kovacsgabor:
This reverts commit 8c33579622ea4a2d16e1587f487fd4a7cb9f5d30.

Resolves #322.


kovacsgabor:
@nboldi, please review.

kovacsgabor:
@ELTE-Soft/all, could someone review this pull request? This is a very small update.

djnemeth:
The changes are reasonable, and it seems to be working. I think this pull request is ready to be merged.

PR #324, opened 2016-04-12
Renamed hu.elte.txtuml.export.uml2.tests.models_ as hu.elte.txtuml.export.uml2.tests.models

### LINKED ISSUE #290: Rename hu.elte.txtuml.export.uml2.tests.models_
Remove the underscore from the end.

kovacsgabor:
Resolves #290.


kovacsgabor:
@nboldi, please review.

gerazo:
As far as I know, Boldi used it because of technical reasons. Now I see that it is working even without the underscore. (This is kinda similar with what I am fighting now.)

kovacsgabor:
Yes, he told me that the name without the underscore caused some problems before. I do not know what kind of problems were these, but everything seems to work now.

nboldi:
If all unit tests pass then the problem was most likely in my configuration.

Issue #219, opened 2016-01-27
Remove pingpong example


kovacsgabor:
The ```pingpong``` example is very messy. As soon as we have a better one which uses all the port and connector features included in this one, ```pingpong``` should be removed.

kovacsgabor:
@devaigergely81, what do you think about this proposal?

I find the `clock` example a lot better one that could replace `pingpong`, although it does not contain any delegation connectors.

devaigergely81:
We could extend the clock example with delegation connectors:
- The `clock` class could have a port that accept signals to reset the hour, minute and second values. So delegation connectors would go to the three roles.
- Another port could be connected to the `display` to send the time strings out.

kovacsgabor:
After the refactoring in #326, the removal of this example is unnecessary.

PR #326, opened 2016-04-12
Refactored the pingpong demo

kovacsgabor:
The pingpong example was very messy (as stated in #219) and therefore needed a major refactoring.

It was not the proposed solution in #219 but it makes that issue invalid.


examples/demo/pingpong/src/pingpong/j/model/associations/LeftPlayer.java
```diff
@@ -0,0 +1,13 @@
+package pingpong.j.model.associations;
+
+import hu.elte.txtuml.api.model.Composition;
+import pingpong.j.model.Player;
+import pingpong.j.model.Game;
+
+public class LeftPlayer extends Composition {
+	public class inGame extends HiddenContainer<Game> {
```
gerazo: Why isn't it capital? Same true for player and PlayerOwnsRacket.player and .racket and for RightPlayer.

gerazo:
Otherwise the above comment I like it, the names are much more descriptive than before.

kovacsgabor:
We used to give non-capital names for association ends and connectors. This is against Java conventions but it makes the code more easily readable in my opinion.

Because this way instead of
```java
Action.link(LeftPlayer.InGame.class, game, LeftPlayer.Player.class, player);
```
we write
```java
Action.link(LeftPlayer.inGame.class, game, LeftPlayer.player.class, player);
```
which makes it clearer that this is the name of an association and its end.

gerazo:
Ok, fair enough, let's merge!

PR #331, opened 2016-04-17
Model execution

### LINKED ISSUE #16: Reduce reflection in txtUML runtime lib

### LINKED ISSUE #62: Multiple model executions at runtime
By making the ```ModelExecutor``` class instantiatable, it could be possible to run more than one model execution at once (on different threads). This would also bring our implementation closer to a parallel execution. However, this would have some overhead as certain ```Action``` calls (which should remain static for the convenience of the user) would need to access the ```ModelExecutor``` instance which runs them. For this, the ```Thread.currentThread``` method could be used (to get the current ```ModelExecutorThread``` and through it the ```ModelExecutor``` instance) which is certainly slower than a static method call.

### LINKED ISSUE #63: Model execution initialization
In case of a running JtxtUML model, sending signals to model objects from an outer source (from code running on a different thread than the actual model executor thread) is always safe, that is, it cannot break the execution.

Certain other operations (like creating and linking new objects) can also be performed safely, but only with strict restrictions (to maintain thread-safety) which are not so self-evident. However, in a well-designed model, these operations are only needed to initialize the model execution (to populate the model).

If these initializing operations could be done **before** the model execution starts, than the same rules would apply to them as at any point during the execution (from inside the model). It would be even safer (or at least, much clearer) if these operations could be done on the model executor thread.

My proposal is to implement a ```ModelExecutor.start``` static method (or non-static with #62 ) which would accept an instance of a ```Configuration``` (or ```Initialization```) interface and would run its only method at the beginning of a model execution. Of course, this instance could be created with a lambda expression to be short.

This way, we could have a simple rule: from outside the model, only sending signals to objects asynchronously (with ```Action.send```) is legal, all other actions are prohibited to ensure the safety of the execution.

### LINKED ISSUE #72: Standard library as one or more models?

### LINKED ISSUE #131: The runtime API lacks proper error handling

### LINKED ISSUE #170: Model object life cycle
Review the life cycle of model objects in the API. Consider introducing final states.

### LINKED ISSUE #255: Separate model execution
This feature is already under development and will introduce a new model execution mechanism which will be fully separated from the modeling API (communicating with it only through specific interfaces and abstract classes), resulting in clearer, more easily maintainable code and making it possible to create different model execution engines.

This is a major update which will also solve the following issues: #16, #62, #63, #131, #170 and hopefully #254.

kovacsgabor:
This pull request resolves #255 and as it includes a major refactoring of the modeling API, it also
* resolves #16,
* resolves #62,
* resolves #63,
* resolves #72,
* resolves #131,
* resolves #170.


kovacsgabor:
retest this please

kovacsgabor:
@gerazo, could you take a look at this test failure? I cannot understand what is the problem of Jenkins here.

gerazo:
retest this please

kovacsgabor:
retest this please

kovacsgabor:
retest this please

gerazo:
retest this please

gerazo:
retest this please

gerazo:
retest this please

gerazo:
retest this please

kovacsgabor:
This a heavy pull request so I do not want to explicitly name someone, but @ELTE-Soft/all please, review!

kovacsgabor:
@nboldi, could you review this pull request?

nboldi:
Yes, I'll review it soon.

nboldi:
Maybe `Action` could be part of `hu.elte.txtuml.api.model.execution`, because it contains execution-related code.

nboldi:
A note: The example project might be needed to manually refresh. To do this remove the library dependency and add it again (Configure Build Path / Libraries / Add Library / txtUML model execution runtime)

nboldi:
Jenkins, retest this please

kovacsgabor:
`Action` is called from model code and therefore it must be part of `api.model`.

nboldi:
OK. I found no other problem, so merge can precede.

PR #332, opened 2016-04-18
Fix/issue325

### LINKED ISSUE #325: In case of a missing initial transition, an uninformative warning message is shown
If a class has an initial pseudostate but no initial transitions, then the executor shows a warning about an arrived but unprocessed `null` signal at the object. This is in no way informative for the user.

kovacsgabor:
Fixes #325.


kovacsgabor:
@djnemeth, please review.

djnemeth:
Seems to be working. It's nice that fully qualified names are used in error messages, they are more informative, especially in case of hierarchical state machines.

I suggest to extend this (in a separate pull request, of course) to [execution logging](https://github.com/ELTE-Soft/txtUML/blob/dd66726299beebbefad1dc001e85a5db9c8479d9/dev/plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/ModelExecutor.java#L136)  in some form, at least with class-qualified names (e.g. `MyClass.MyCompositeState.MyInitialState`)

PR #337, opened 2016-04-21
Changed model declaration

### LINKED ISSUE #217: Rename model keyword in XtxtUML?
Currently, the keyword ```model``` is used in ```model-info.xtxtuml``` to set the name of the model. This way, however, the word ```model``` has to be written as ```^model``` in any XtxtUML source which is not a problem in itself but as the qualified name of the JtxtUML API packages contain this word, all imports that reference the API appear as ```import hu.elte.txtuml.api.^model.``` which is not really nice.

### LINKED ISSUE #328: Change syntax of model declarations?
In the current implementation, the XtxtUML runtime depends on components of the Xtext UI. See #320. One notable example for this dependency is [XtxtUMLPackageNameCalculator](https://github.com/ELTE-Soft/txtUML/blob/0d885491df573c697b7139a1d5779aff9e5510e0/dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/naming/XtxtUMLPackageNameCalculator.xtend#L14), which is [used](https://github.com/ELTE-Soft/txtUML/blob/0d885491df573c697b7139a1d5779aff9e5510e0/dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/jvmmodel/XtxtUMLJvmModelInferrer.xtend#L69) by the model inferrer to calculate the fully qualified name of the JVM element representing a model declaration, from which a *model-info.xtxtuml* file will be generated, eventually translating to a *package-info.java* file.

The current syntax of model declarations is the following:

```
'model' name=STRING? ';'?
```

That is, a `model` keyword followed by an optional string literal representing the name of the model (which is used during UML export if I recall correctly) and an optional semicolon at the end. It can be seen that no fully qualified name is specified.

In XtxtUML files containing a package declaration (as in Java), the package name is actually the name of the whole file as a container. This is leveraged by the default qualified name calculator, which computes the qualified name of an XtxtUML element by traveling through its containers and appending their names, yielding a fully qualified name if a package declaration is present.

The easiest way to eliminate the aforementioned problematic dependency would be to change the syntax of model declarations such that the package name is included as well. In this way, the default qualified name calculator could be used instead of the UI-specific package name calculator. This would also bring [*J*](https://github.com/ELTE-Soft/txtUML/blob/db9edb14ee264aa6c5c92c68f08962fcab1feb37/examples/demo/clock/src/clock/j/model/package-info.java) and *X* syntaxes closer.

I propose the following alternatives:

1. alternative
<pre>
    <b>model</b> foo.bar.baz <b>named</b> "BazModel";
</pre>

2. alternative
<pre>
    <b>model</b> foo.bar.baz <b>as</b> "BazModel";
</pre>

3. alternative
<pre>
    <b>model</b> "BazModel";
    <b>package</b> foo.bar.baz;
        <i>// or in reverse order</i>
</pre>

4. alternative
<pre>
    <b>model</b> foo.bar.baz {
        <b>name</b> "BazModel";
    }
</pre>

5. alternative
<pre>
    <b>model</b> "BazModel" {
        <b>package</b> foo.bar.baz;
    }
</pre>

#217 should be considered too. @ELTE-Soft/all, what do you think?

kovacsgabor:
To resolve #328. Also resolves #217.


kovacsgabor:
@djnemeth, please review.

djnemeth:
Looks good to me! :+1:

kovacsgabor:
Do you agree with this last commit, @djnemeth?

kovacsgabor:
@djnemeth, this branch fails now, because the validations merged to master in #319 use the custom package name calculator that is removed in this branch. I could revert the removal but it was the aim of #328 (as far as I understood) to remove that custom calculator.

What do you think about this?

djnemeth:
I was already merging while you pushed your merge commit. Indeed, merging this pull request first would be the better option, sorry, I messed up. I suggest to merge with removing the validation of package names, and re-introduce it later in as a UI validation in a separate pull request.

djnemeth:
I have a maven-verified merge commit in my local repository, ready to be pushed. Is it possible to **reset** your last merge commit to make the history a bit cleaner?

kovacsgabor:
Of course, but you won't like my solution for this situation...

kovacsgabor:
Or you **meant** to solve it with hard reset and force push? Because that is what I will do.

djnemeth:
Yes, I meant exactly that. ( ͡° ͜ʖ ͡°)

kovacsgabor:
I'll merge this branch if the check is completed successfully. Do you agree, Dávid?

djnemeth:
Absolutely.

Issue #188, opened 2015-12-03
Dependency cycles caused by test projects


kovacsgabor:
When exporting txtUML as a Deployable feature with all test projects imported in the workspace, the exportation fails because of one of the following two dependency cycles it finds:

hu.elte.txtuml.api.model
hu.elte.txtuml.api.stdlib

and

hu.elte.txtuml.xtxtuml
hu.elte.txtuml.xtxtuml.ui

kovacsgabor:
To sum up our conversations about this matter: Our test projects are now fragments of the tested projects, and as dependencies added to the fragments, the host projects are effectively modified as well.

So in the case of the ```api.model``` and ```api.stdlib``` projects, as the latter depend of the former, and the test project ```api.model.tests``` is a fragment of ```api.model``` with explicit dependency on ```api.stdlib```, the cycle is complete. The solution to this situation would be the move every test from ```api.model.tests``` project which depend on ```api.stdlib``` to a new ```api.stdlib.tests``` project and remove the ```api.model.tests``` -> ```api.stdlib``` dependency.

In case of the ```xtxtuml``` and ```xtxtuml.ui``` projects, the problem is the same, and therefore a similar solution should solve it.

PR #338, opened 2016-04-21
Removed unncessary dependency of xtxtuml.tests on xtxtuml.ui

kovacsgabor:
Fixes half of #188.

The other half is fixed by #331.


kovacsgabor:
@djnemeth, I do not see the point of this dependency, but the Jenkins build fails. De you know the reason?

djnemeth:
It seems the standard test project created by Xtext is intended for both runtime and UI testing. We have to introduce separate test projects.

As I'm planning to implement a comprehensive unit test suite for all XtxtUML components in the upcoming weeks — as part of the finalization process of my BSc thesis work —, I suggest to close this pull request for now.

Issue #181, opened 2015-12-03
No wizard to create XtxtUML files


kovacsgabor:
Currently, we have no wizard to create files with extension ```.xtxtuml``` other than a ```model-info.xtxtuml```.

nboldi:
There should be wizards for model class, association, signal, etc...

devaigergely81:
For the moment, I have written in the documentation how to create model elements by creating a new file with appropriate extension: http://txtuml.inf.elte.hu/wiki/doku.php?id=v030:userguide#new_model_elements

We have to think about the wizards proposed above, because the users will not always want to create a new file for each model element...

Issue #184, opened 2015-12-03
XtxtUML as default language in 'New model' wizard


devaigergely81:

devaigergely81:
I agree with this suggestion originally proposed by @kovacsgabor in #171. What do @ELTE-Soft/all think?

nboldi:
I think it is a management question. Which technology is the primary product? Is XtxtUML just a nice syntactic sugar? Is JtxtUML a replacement if something goes wrong with the toolchain?

dobreffandras:
I have the same question as @nboldi "Which technology is the primary product?"
Do we want to set default the one which is more convenient to the user (X) ?
(As I see once the user had tried out the DSL, won't write models in the J syntax)
Or do we want to emphasize that our product is a modelling tool in Java with an optional syntax?

devaigergely81:
In an Eclipse wizard the X syntax is a safe default option. The Java variant is important outside Eclipse: I know about an upcoming project where someone wants to *generate* txtUML models and run them automatically (headless Eclipse or without Eclipse) for simulation purposes. In this case the J syntax seems to be a better technical solution.

gerazo:
Unfortunately, without measuring on a broad enough, independent user-base, we cannot surely tell if X or J is better. Our opinion as developers does not count, only real use-cases, user interviews.

kovacsgabor:
The team decided to support this proposal for now.

PR #341, opened 2016-04-24
'New XtxtUML File' wizard and other wizard-related improvements

djnemeth:
For a brief overview, see the (full) commit messages.

The new wizard:
new-x-file


djnemeth:
We should discuss whether this actually resolves #181. If the team agrees on #184, it could also be implemented within this pull request.

djnemeth:

dobreffandras:
Nice refactorings. 
I think this resolves #181, but we should open a new topic for creating wizards for model elements. It's a little bit confusing for the first time that the 3 thing that can be created are: `txtUML Project`, `txtUML Model`, and `XtxtUML File`, so a discussion is needed.

Smoke test: During the creation of a XtxtUML file in a J model, eclipse asks if I want to add Xtext nature  to my project. (First question: Should the user be asked this question?) Hitting `Cancel` or `X` creates the files and does not add the nature. This behavior is expected hitting `No`, after hitting `Cancel` or `X` I expect the file not to be created and te wizard stay on screen.

I would wait with the merge until #184 is resolved.

PR #342, opened 2016-04-25
Changed 'getSignal' to 'getTrigger', 'sigdata' to 'trigger'

### LINKED ISSUE #335: Trigger/getTrigger?
We agreed in #166 to change the XtxtUML keyword `sigdata` to `signal` (#333 is about to implement this change) which sounded pretty well on its own, but as I look at the following code in this proposed syntax, I find it pretty hard to read:
```
transition StartGame {
	from Waiting;
	to Waiting;
	trigger Ball;
	effect {
		Player left = this->(LeftPlayer.player).selectAny();
		send signal to left;		
	}
}
```
It simply sends forward the received signal, but when I read it, I ask myself "We are sending here... which signal? Any signal? What is the type of this signal?" because it seems as a signal definition not a query... This assumption is even more confirmed given the fact that the `signal` keyword is used to define new signals.

My proposal is to use the `trigger` keyword in X and the `getTrigger` method in J as I think this would tell the reader exactly what is happening here. We read that the transition is triggered by signals which are of type `Ball` and then in the effect an instance of `Ball` is queried with the `trigger` keyword. Moreover, if we ever want to use any other kind of events in the language which can be triggers, the trigger/getTrigger pair would still be usable.

@ELTE-Soft/all, what do you think?

Note: this is a long discussion for a very simple change in the language as it only affects the names of some keywords and methods. It can be implemented in minutes.

kovacsgabor:
To resolve #335.

### DIFF
--- a/dev/plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/Region.java
+++ b/dev/plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/Region.java
@@ -324,7 +324,7 @@ public class Region extends StateMachine {
 	 */
 	private void executeTransition(Transition transition) {
 		callExitAction(transition.getSource(),
-				transition.getSignal(Signal.class));
+				transition.getTrigger(Signal.class));
 		Report.event.forEach(x -> x.usingTransition(this, transition));
 		transition.effect();
 		currentVertex = transition.getTarget();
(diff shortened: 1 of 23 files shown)


kovacsgabor:
@djnemeth, please review.

djnemeth:
Looks good to me. :shipit:

PR #345, opened 2016-04-27
Uml export refactor

### LINKED ISSUE #237: Constructors should be labeled as "Create" standard stereotype
CPP exporter can't generate correct constructors without knowing that an operation is indeed a constructor.

### LINKED ISSUE #260: Refactor uml export

### LINKED ISSUE #305: Change request about the export of if-then-else
~~~~
class A {
	void f() {
		int i;
		if(i == 0) {
			i = 2;
		} else {
			i = 4;
		}
	}
}
~~~~
ifproblem

An if-then-else is exported as a conditional node with two clauses. Each clause has a *test*, which is responsible for evaluating the condition of the clause and a decider (an output pin) that specifies the result of the computation. Since we want to avoid duplicate condition evaluations, the test of the then clause writes the result into a variable, which is then negated in the test of the else clause.

This causes data dependency between the two tests, which is represented by the object flow highlighted in the picture above. When determining the correct order in which the tests should be generated into the C++ code, this data dependency needs to be analysed. As a general design decision, we want the exported models to be explicitly structured, so that such data dependency analysis is never required. (That might be later extension.)

Suggestion: Use one single test for all clauses, that creates one boolean local variable per clause, writes them (to set the values) and reads them (to provide output pins for the deciders).

Another request: the names of these variables should be Java/C++ identifiers, the # symbol currently causes problems.

In case an if- else if - else if - .... - else structure, currently multiple conditional nodes are generated embedded in each other. It would be nicer to have one single conditional node with many clauses. But this is not a blocking issue.

### LINKED ISSUE #308: Constructor call when create a Signal
``
public class S1 extends Signal { 
public S1(int i) {
this.i = i;
}
}
``
``
S1 s = new S1(5);
``

singal_error
Although signal `S1` has a constructor parameterized with an integer, it seems that there is no a CallOperationAction which invokes it.

### LINKED ISSUE #315: Foreach loop exported wrong order
foreach_problem

The `ReadLinkAction` activity should be prevented the `ExpansionRegion` because I need its result for the iteration.

### LINKED ISSUE #317: LinkAction problems
- The `ReadLinkAction` should contain one more `LinkEndData` whose `End` defines the read property and it contains no value.
- It should be defined as a standard function for `selectAny()` whose parameter contains the result of `ReadLinkAction` and it returns only one object.
- The `CreateLinkAction` contains two unnecessary `ValuePin`s.

### LINKED ISSUE #318: Composition is not appearing on the generated Diagram
Compositions in the model are not appearing on the diagram

```java
class ExampleComposition extends Composition {
	  class a extends Container<A> {}
	  class b extends Many<B> {}
}
```
Result can be seen in this picture:
composition

### LINKED ISSUE #329: Unnecessary opaque action causes problems
opaque_action_problem

The highlited `InputPin` should be linked with the result of `ReadStructuralFeatureAction` instead of the unknown `OpaqueAction`. This model was written in xtxtuml so this problem might be caused by the generated code because the `ValueOf..` statements are not valid in the model.

nboldi:
Completely redesigned the UML export module. Export is one-stage, cached and compositional. Uniform export for structure and action code.

Solves the following issues: 
* resolves #329, 
* resolves #318,
* resolves #317,
* resolves #315,
* resolves #308,
* resolves #305,
* resolves #260,
* resolves #237.

Much better than the old version, but there are some missing features and known errors in the refactored version:
 - Call of super default constructor at the beginning of a constructor, when there is no explicit call.
 - The types and parameters concerning signal constructors are not entirely correct
 - Connect and send actions for ports are not supported (couldn't find the correct UML representation)
 - `selectAny` is implemented as an StdLib operation, that is not advantageous
 - Convertion of values to String is not supported. This includes automatic conversion as well as specific calls (eg: `Boolean.valueOf(...)`).
 - The performance is not quite good. Profiling showed that the reason is that loading elements from the `ResourceSet` is slow. This could be improved with caching and/or parallel loading.


kovacsgabor:
Note: I prefaced each issue reference with the "resolves" keyword in the initial message because of https://help.github.com/articles/closing-issues-via-commit-messages/

kovacsgabor:
retest this please

gerazo:
retest this please

kovacsgabor:
This branch cannot export X models because it does not find the files in the src-gen folder.

kovacsgabor:
Exportation of the pingpong example fails.

nboldi:
Yes it fails because action code for port handling is not (yet) supported by uml export, and things like OutPort and Interface.Empty are also unsupported.

kovacsgabor:
But this way we are loosing features, because the pingpong could be exported before the refactoring.

I suggest a quick fix for this: when the export finds anything in the action code that it does not know, it should put an OpaqueAction there which shows that the certain part cannot be exported. And at the end of the export, a warning should be shown.

nboldi:
No, only the original pingpong example could be exported, the rewritten couldn't be.

nagyandras95:
As I see there are some more missing feature in this branch:
- the composite states are generated as simple states
- it cannot export the choice pseudo states, the export fails on them (however the old version could export them as I remember) 
- it is unknown what node the return statement is in case of guards

But I think these issues are not so important, all demo example can be exported in non pedantic mode and its original purpose was only the refactoring that seems to be correct  so you should merge from master and try if the new test are passed on this branch.

kovacsgabor:
It would be nice to export choice pseudostates as well.

nboldi:
But that is not in scope of the refactoring.

kovacsgabor:
They were exported by the old version of the exporter so they should be exported by this new version as well.

nboldi:
OK. Implemented choice states and composite states for backward compatibility.

nboldi:
@gerazo, @nagyandras95 : Please check which C++ test fails and resolve it.

gerazo:
Boldi, all of them fails. In fact, not the CPP compilation fails but sooner the export itself fails. Although the other export only test fails with the same problem. The following is the error result. You can see that after the Eclipse build was called (clean, incremental, auto) and the export starts, it fails with something weird. I am sending you the callstack.

nboldi:
The errors are caused by lack of features in cpp export. This branch should be merged regardless of the resolution of those errors. @gerazo, would you like to ignore the tests for the time being, to prevent master from breaking?

gerazo:
How much would it be to add the missing features (or at least mimic 
somehow)? I wouldn't really like to degrade to a non-compiling CPP 
solution again.

2016-05-12 13:16 keltezéssel, Boldizsár Németh írta:

nagyandras95:
The machine2 model cannot be exported even in error handling mode.

nboldi:
In error handling mode (diagram export) it could be exported, but now I fixed the problem, so it can be exported in C++ code gen mode

nboldi:
I think that the cpp tests are caught in an infinite loop. Jenkins, retest this please.

Issue #235, opened 2016-02-03
cpp testing


gerazo:
How do we write test cases?
Do we compare raw output?
Do we interact somehow with the code?
Are we generating separate mains for test cases?

devaigergely81:
My suggestion:
- Short term: Test cases are txtUML models, the test compiles these to C++ and checks if the generated code is accepted by the C++ compiler.
- Long term: Test cases are extended with main functions which run the generated C++ codes with input data/signals, and output data/signals are compared to predefined values in the test case.
- I **do not** recommend raw output comparison of generated C++ code.

PR #349, opened 2016-05-04
Cpp tests

gerazo:
The tests are finally working. Unfortunately, it turned out that Xtext is buggy and even in headless mode it expects a running xserver and a working window manager. Using its standalone version is not possible when testing as eclipse plugin. So now all txtUML jobs are getting an xvncserver with a real display and a running minimal window manager (metacity) to have all tests working fine. It also means that from now on real UI tests can also be run easily.


gerazo:
retest this please

gerazo:
Solves https://github.com/ELTE-Soft/txtUML/issues/235

gerazo:
Anyone for a review?

gerazo:
A review @nagyandras95 ?

gerazo:
@djnemeth and @nagyandras95 thanks for the review :dog: :dollar:

PR #374, opened 2016-06-14
Fix 371

### LINKED ISSUE #371: Incomprehensible error message when exporting a model in case of missing model declaration
If the package-info.java file is missing from the chosen model package when exporting to UML2, the following error message is shown:
error

### LINKED ISSUE #372: Incomprehensible error message when exporting a model if superpackage is chosen
If I try to export the `clock` example, but instead of the `clock.x.model` package I enter the `clock.x` package as the root package of the model, the following exception is thrown:
error

### LINKED ISSUE #373: Error when exporting the garage example
When exporting the garage example, an error is shown that the /hu.elte.txtuml.examples.garage/gen/hu.elte.txtuml.examples.garage.control.model.uml file is not found.

The error message is valid in the sense that the file is indeed missing, the /hu.elte.txtuml.examples.garage/gen/hu.elte.txtuml.examples.garage.control.model.signals.internal.uml file is generated instead.

I suppose that it is because the hu.elte.txtuml.examples.garage.control.model.signals.internal package (which is an internal package of the model) also contains a package-info.java file (without a `@Model` annotation, which is therefore valid).

nboldi:
Resolves #371
Resolves #373
I hope it also resolves #372


kovacsgabor:
I tested it and all the three issues are indeed resolved.

PR #377, opened 2016-06-20
Fix animation

### LINKED ISSUE #369: Animation fails with EOFException

nboldi:
Fixed the animation to accept file path based URIs.
Resolves #369


kovacsgabor:
The fix of the animation was OK.

The garage example, however, still did not run because it had some bugs in it. I fixed them and now this pull request can be merged.

PR #381, opened 2016-06-27
generate uml into cpp-gen when generating c++ code

nboldi:


nagyandras95:
I think the previous solution was better. The uml model was generated next to the generated cpp files and the runtime library to a model named folder.
cpp-gen
  --model -> uml model
  --runtime
  other files... -> classes, events, etc..
What do you think?

GregoBalu:
The location is now easily adjustable.
The changes look OK to me.

PR #399, opened 2016-07-24
V050 examples prep

### LINKED ISSUE #365: Prepare demo examples for 0.5.0
- [x] All demos should have layout descriptions for all relevant diagrams (both X and J).
- [x] Provide a meaningful naming convention to diagram layout descriptions and configurations which is then followed in all demo examples.
- [ ] Remove warnings (with `@SuppressWarnings("unused")` in case of the airport example where necessary).
- [ ] Mark the airport example as one that is under development (to explain the code parts in comments and `@SupressWarnings("unused")` annotations)
- [x] Rename the airport example to begin with a lowercase letter.
- [x] Rename the 'dokumentacio' folder in the airport example as 'documentation' or remove it entirely.
- [ ] Split the X version of the pingpong and clock examples into multiple files.

GregoBalu:
Resolves #365.
Removed warnings, TODOs, added description in airport.
Split X version of pingpong.


nagyandras95:
Can I eliminate the Collection declarations from the airport model? It is only refactoring, using the links directrly in foreach loops so the generated c++ code might be compilable.

GregoBalu:
You might if that's the only thing blocking the cpp code gen in this model.

nagyandras95:
retest this please

nagyandras95:
There are other problems too so it is not important to eliminate them. We should postpone making airport model compilable.

djnemeth:
I'm not sure if we should move each main model element (classes, signals, etc.) to separate files. Maybe the models would be more readable if we grouped them by types, like in the *airport* example.

I also suggest a more formal and concise description for the *airport* example. Something like
> This example (demo/airport) is under construction. It was made available as its multilayered architecture — where the model defines the business logic of the application — provides a more advanced view on modeling with txtUML.

nagyandras95:
I think it is not problem if they are in separate package like on Machine models or Ping Pong because the package is grouping them by type. Only the producer_consumer model seems to 
unreadable for me because it contains separated files but only one package.

nagyandras95:
The suggested description for airport is more specific so I think it is better.

djnemeth:
> I think it is not a problem if they are in separate packages like in Machine models or Ping Pong because the package is grouping them by type.

Yes, of course, but my concerns are more about the nuisance of browsing through several one-liner files, e.g. signal definitions. I can see that this practice might emerge from the rules of Java (which, of course, can be useful in certain situations), but I believe (supporting that our language allows multiple definitions in one file) it may be reasonable to set a larger scale separation recommendation in our demo examples.

GregoBalu:
I think that it does not matter if it is in one file or more, we need to put this release out asap. These are not critical issues, just aesthetics. We can make a vote on this matter on the meeting but for this release keep it this way and make it ready.

nagyandras95:
I am in similar opinion. The PR should be merged, the release should be launched and we should discuss the architect issues on meeting.

djnemeth:
I started a guest Mars.1 Eclipse to test the split examples, but after importing the demos and several clean-rebuilds I've still got a whole lot of misteriously looking validation errors. I included one from each type below:

> Bound mismatch: The generic method connect(Class<C1>, P1, Class<C2>, P2) of type Action is not applicable for the arguments (Class<Hours.minuteHand>, minuteHand.port(Hand.OutTickPort.class), Class<Hours.hourHand>, hourHand.port(Hand.InTickPort.class)). The inferred type Hand.OutTickPort is not a valid substitute for the bounded parameter <P1 extends ModelClass.Port<I1,I2>>	Clock.java	/clock/src-gen/clock/x/model	line 42	Java Problem
> ...
> Bounds mismatch: The type arguments <Hours.minuteHand, Hand.OutTickPort, Hours.hourHand, Hand.InTickPort, Empty, TickIfc> are not a valid substitute for the bounded type parameters <C1 extends ConnectorBase.ConnectorEnd<?, P1>, P1 extends ModelClass.Port<I1, I2>, C2 extends ConnectorBase.ConnectorEnd<?, P2>, P2 extends ModelClass.Port<I2, I1>, I1 extends Interface, I2 extends Interface> of the method connect(Class<C1>, P1, Class<C2>, P2)	clock.xtxtuml	/clock/src/clock/x/model	line: 37 /clock/src/clock/x/model/clock.xtxtuml	XtxtUML Problem
> ...
> The method send(S, Reception&lt;S>) in the type Action is not applicable for the arguments (HandValue, Hand.this.port(Hand.ValuePort.class).required::reception)	Hand.java	/clock/src-gen/clock/x/model	line 59	Java Problem
> ...
> The type TickIfc does not define reception(S) that is applicable here	Hand.java	/clock/src-gen/clock/x/model	line 57	Java Problem
> ...
> Type mismatch: cannot convert from Ball to Ball	racket.xtxtuml	/pingpong/src/pingpong/x/model	line: 43 /pingpong/src/pingpong/x/model/racket.xtxtuml	XtxtUML Problem

Basically, every `connect` call is erroneous in the *pingpong* example, and also there's the infamous *cannot convert from A to A* error (which I think might be somehow related to the now partially reverted #89). `connect` calls are also erroneous in the *clock* example, but in addition, even the generated files contain errors about missing receptions (which are defined correctly in corresponding Java files), whereas their XtxtUML sources don't!

This is *really* strange. Do you experience anything listed above?

djnemeth:
It seems that the problem was on my side: I simply used `git checkout` to switch between branches, but the *src-gen* folders of the demo projects hadn't been cleaned automatically. This resulted in the cryptic errors mentioned above.

I tested the modified examples, they seem to be working. Let's merge this!

PR #400, opened 2016-07-25
List empty packages too in txtUML wizards

### LINKED ISSUE #393: XtxtUML models with subpackages are not browsable in txtUML wizards
Steps to reproduce:

1. Create a new txtUML project.
2. Create a new XtxtUML model inside the previously created project.
3. Create a new (sub)package inside the previously created model.
4. Open either the *Generate C++ code from txtUML* or the *Generate diagrams from txtUML* wizard, select the project created in step 1 and try to select the model created in step 2 by the *Browse...* option. Notice that it's not listed.

For an additional example, consider the following package structure:
```
model
├── foo
├── bar
|   └── foo
└── foobar
    └── foo
        └── bar
```
In this case, packages `model.foo`, `model.bar.foo` and `model.foobar.foo.bar` are selectable as models, whereas package `model` (the actual model) is not.

djnemeth:
See the (full) commit messages for details. This (hopefully) resolves #393.

### DIFF
--- a/dev/plugins/hu.elte.txtuml.export.cpp/src/hu/elte/txtuml/export/cpp/wizardz/TxtUMLToCppPage.java
+++ b/dev/plugins/hu.elte.txtuml.export.cpp/src/hu/elte/txtuml/export/cpp/wizardz/TxtUMLToCppPage.java
@@ -130,8 +130,7 @@ public class TxtUMLToCppPage extends WizardPage {
 				} catch (JavaModelException | NotFoundException ex) {
 				}
 				PackageSelectionDialog dialog = new PackageSelectionDialog(getShell(), getContainer(),
-						PackageSelectionDialog.F_HIDE_DEFAULT_PACKAGE | PackageSelectionDialog.F_REMOVE_DUPLICATES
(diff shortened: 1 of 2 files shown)


djnemeth:

GregoBalu:
Seems OK to me.

nagyandras95:
Is it not problem that the dialog will list empty packages too? It can be annoying for an user especally beacause of packages names like this: hu.elte.txtuml.some.model.. :) In this way the hu and elte .. etc will be listing too. However it is better than not listing a model package at all.

GregoBalu:
I believe that it's already filtered that only `@Model` annotated packages will be listed here.

djnemeth:
As far as I know, model packages are *not* filtered currently. If this is true, I recommend the following alternative: for now, merge this pull request (as a bugfix), and add model filtering later, in a separate pull request (as a convenience feature).

djnemeth:
The team have agreed that this PR should be merged for now, and proper filtering should be implemented later.

nagyandras95:
Ok, then it can be merged this PR.

Issue #394, opened 2016-07-21
Can't generate class diagrams from XtxtUML models


djnemeth:
Steps to reproduce:

1. Open the *Generate diagrams from txtUML* wizard.
2. Select any of the demo example projects.
3. Select the corresponding XtxtUML model.
4. Add the corresponding class diagram description (*X«DemoName»ClassDiagram*).
5. Click on *Finish* and notice that the opened Papyrus diagram is empty.

It seems this issue occurs only in case of class diagrams of XtxtUML models. JtxtUML models and diagrams are apparently unaffected, and state machine diagrams from XtxtUML models can also be visualized.

dobreffandras:
I've made some debugging and realised that the model itself is not included in the txtUML-Uml2 mapping  (saved to the .mapping file).

PR #401, opened 2016-07-25
Fixed the model cache to generate the model with the current name.

nboldi:


djnemeth:
@nboldi, please specify the context of this pull request (briefly, for better understandability and traceability).

nboldi:
The names that are used as keys in the model cache are used when the mapping for the model is exported for diagram editing. To fully work with models that are created in the X syntax, I had to fetch the same package in multiple source folders. The keys were the package names, but they had been concatenated. That caused problems for diagram export.

dobreffandras:
Seems to be working now (Mars2, mirror target, runtime Eclipse). Thanks @nboldi

dobreffandras:
resolves #394

PR #416, opened 2016-08-01
Fixed the export of the + operator to only export the concat operatio…

### LINKED ISSUE #409: Numeric expression is accidentally converted to string
The generated code for the expression `world.v * (world.h*world.h/500000 + world.h/500 - 2)` is
```
int temp3 = (world_us0->h * world_us0->h);
int temp4 = (temp3 / 500000);
std::string temp5 = conversion::to_string(temp4);
int temp6 = (world_us0->h / 500);
std::string temp7 = conversion::to_string(temp6);
std::string temp8 = concat(temp5,temp7);
int temp9 = (temp8 - 2);
```
It seems that the `+` operator is used as concatenation.

nboldi:
…n if Strings are concatenated.

Resolves #409


nagyandras95:
Seems to OK.

PR #417, opened 2016-08-01
Removed unsupported files from demos

GregoBalu:
Resolves #414  and #412 .

PR #418, opened 2016-08-01
Easy solution for the missing subpackages in UML export. Only the sou…

nboldi:
…rce dir is exported where package-info.java is found.


nboldi:
The fix supports multiple X and J models in a project, but not a model that is composed of both X and J files.

nagyandras95:
I have tested to some x model, it seems to be a good solution for release.

djnemeth:
I have tested this in a guest Eclipse by generating every demo diagram we provide a layout definition for (excluding composite structure diagrams, see #417). Now, all of them is generated properly. :+1:

Issue #320, opened 2016-04-11
XtxtUML runtime depends on UI components


djnemeth:
In the current implementation, the XtxtUML runtime [depends](https://github.com/ELTE-Soft/txtUML/blob/db9edb14ee264aa6c5c92c68f08962fcab1feb37/dev/plugins/hu.elte.txtuml.xtxtuml/META-INF/MANIFEST.MF#L23) on components of the Xtext UI module. As far as I know, this makes headless builds (e.g. unit testing such UI-dependent classes) impossible. @gerazo, please correct me if I'm wrong.

**Edit:** Corrected *XtxtUML UI module* to *Xtext UI module*.

gerazo:
You are right. However, I can't see hu.elte.txtuml.xtxtuml.ui in the manifest of hu.elte.txtuml.xtxtuml. I only see dependency in the pom.xml but it is for building reasons (cleaning the generated folder synchronuously between the two projects). I am probably missing something.

djnemeth:
Sorry, you're right. It [depends](https://github.com/ELTE-Soft/txtUML/blob/master/dev/plugins/hu.elte.txtuml.xtxtuml/META-INF/MANIFEST.MF#L23) on UI components of *Xtext* itself.

gerazo:
I've checked and the Xtext generated manifest does not depend on these UI components, so we added them, so it's our fault. :facepunch:

PR #427, opened 2016-08-13
Fixes and improvements in validation

### LINKED ISSUE #29: Check java reserved names
`state final;` is a perfect state declaration in XtxtUML, but the generated code is bogus:
```
public class final extends StateMachine.State {}
```

### LINKED ISSUE #40: Collection support needed, confusing error message
```
Collection<PerformanceTestModel.A> a = this->PerformanceTestModel.AB.a;
```

This instruction produces the following error:
```
Type mismatch: cannot convert from Collection<PerformanceTestModel.A> to Collection<PerformanceTestModel.A>
```
Maybe a java type should not be used that way, but then what is the suggested method to deal with collection types?

### LINKED ISSUE #343: Association not accessible from inheriting class
inheritassociationbug

### LINKED ISSUE #387: Validation of package and model declarations is needed
Currently it's not validated whether declared package names (i.e. in package and model declarations) correspond to expected ones (considering the underlying folder structure). As named language entities (e.g. classes) get generated according to their fully qualified name — which is based on the _declared_ package name in Xtext —, inconsistent package declarations result in misplaced generated elements. This can cause unexpected and hardly localizable linking errors. Therefore I believe such validation is highly needed.

### LINKED ISSUE #388: Uniqueness validation of class-like elements should be case insensitive
Consider the following points:

- XtxtUML is case sensitive.
- Certain XtxtUML elements are inferred as Java (nested) classes (e.g. classes, signals, states, transitions...).
- The Java compiler generates a *.class* file from each (nested) class definition.
- The Windows API is case *in*sensitive.

As a result, if the fully qualified names of two class-like XtxtUML entities differ only in case, the Java compilation fails on Windows. As far as I know, Mac OS is affected as well. See the following example:

case_insensitivity_error

Uniqueness validation of class-like elements should therefore be case insensitive.

### LINKED ISSUE #390: Uniqueness validation of connector ends should be revised
In the current implementation, [uniqueness validation of connector ends](https://github.com/ELTE-Soft/txtUML/blob/c94570311efb3ca1a1c8ee728cc0b6ef2daf912c/dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/validation/XtxtUMLUniquenessValidator.xtend#L242) disallows left-reflexive connectors — i.e. connectors with ends specifying the same *role* —, but allows right-reflexive connectors — i.e. connectors with ends specifying the same *port* —, provided that the reflexively-connected port is compatible with itself. In case of assembly connectors, self-compatibility holds if and only if both the provided and required interface of the given port is the same. In case of delegation connectors, a port is always compatible with itself. See *[XtxtUMLConnectorValidator.xtend:76](https://github.com/ELTE-Soft/txtUML/blob/c94570311efb3ca1a1c8ee728cc0b6ef2daf912c/dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/validation/XtxtUMLConnectorValidator.xtend#L76)*.

Partially based on our latest meeting, contrary to existing validation rules:
- in case of *assembly* connectors:
  - role duplication should be *allowed* (this might be reasonable while implementing e.g. a feedback-based system)
  - port duplication should be *disallowed**

- in case of *delegation* connectors:
  - role duplication should be *disallowed* (role duplication would violate the rule that each delegation connector should connect two *different* roles of the same composition)**
  - port duplication should be *allowed* (e.g. in case of a reflexive underlying composition)***

\*: Are we sure? In our meeting, we based this decision on the presumption that a port would be compatible itself via an assembly connector if and only if both its provided and required interface was empty, but I believe compatibility holds in a more general case as well, see above. However, if we allow *both*\*\*\*\* role and port duplication in case of assembly connectors, we will have to ensure that a port instance doesn't get linked with itself, as this would result in an infinite loop (or *worse*) when a signal arrives, as far as I know.

\*\*: Currently the [validation method](https://github.com/ELTE-Soft/txtUML/blob/c94570311efb3ca1a1c8ee728cc0b6ef2daf912c/dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/validation/XtxtUMLConnectorValidator.xtend#L44) responsible for checking underlying compositions rely on uniqueness validation to ensure that the roles are different. We might need to change this too.

\*\*\*: Reflexive compositions *are* valid in UML, right? Moreover, the runtime validation rule — mentioned in \* — about connecting the same port instances are needed here as well. This kind of runtime validation should be placed inside `Action.connect`, if it hasn't been implemented yet.

\*\*\*\*: The main problem we have to avoid here is that a port instance cannot be connected with itself. An interesting question: is it enough if the roles specified in the given connector are different (even if the ports are the same) to guarantee that this could not happen? This question boils down to whether (in UML) the same instance can be linked via two compositions simultaneously, provided the container instance is the same?

### LINKED ISSUE #426: Strangely behaving JtxtUML error markers in generated files
Steps to reproduce:

1. Comment out line [XtxtUMLDerivedResourceMarkerCopier:114](https://github.com/ELTE-Soft/txtUML/blob/52b00ef7e4402e7ab9382a9c229b025b3488969d/dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/validation/XtxtUMLDerivedResourceMarkerCopier.xtend#L114). This is to make sure original JtxtUML error markers don't get deleted by the XtxtUML error marker propagator.
2. Change line [Utils:36](https://github.com/ELTE-Soft/txtUML/blob/52b00ef7e4402e7ab9382a9c229b025b3488969d/dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/visitors/Utils.java#L36) by removing ` || modifier.isProtected()`. This is to introduce an easily checkable JtxtUML validation error.
3. Launch a guest Eclipse, then create a new txtUML project with a new XtxtUML model.
4. Create a new XtxtUML file inside this model, and define the following class in it:

  ```Java
  class A {
      public int a;
  }
  ```
5. Save the XtxtUML file.
6. Change the `public` modifier to `protected` and save the XtxtUML file.
7. Notice that although the JtxtUML validator should mark the generated *A.java* file invalid (see step 2), it appears to be valid.
8. Open the generated *A.java* file. Notice that the missing error marker is displayed now.
9. Close the generated *A.java* file. Repeat the procedure from step 4 (you don't have to create a new file though) to step 6. Instead of opening the generated file, initiate a full clean & rebuild. Notice that in this way, the missing error marker appears (duplicated, see step 1).
10. Open the generated *A.java* file. Repeat the procedure from step 4 (you don't have to create a new file though) to step 6. Notice that in this way, the missing error marker appears, but the error doesn't get propagated to the XtxtUML file.

I believe this strange behavior is related to the fact that the [JtxtUMLCompilationParticipant](https://github.com/ELTE-Soft/txtUML/blob/52b00ef7e4402e7ab9382a9c229b025b3488969d/dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/JtxtUMLCompilationParticipant.java) is [hooked](https://github.com/ELTE-Soft/txtUML/blob/52b00ef7e4402e7ab9382a9c229b025b3488969d/dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/JtxtUMLCompilationParticipant.java#L58) to the *start* of the build and not the *end* of it. As a result, I think it might happen that JtxtUML error markers get removed during the *clean* phase which occurs at the start of Java builds.

djnemeth:
This pull request is mainly intended to fix various validation-related issues. In addition, some aspects of both the J- and XtxtUML validator has been improved. For more information, see the (full) commit messages.

**Some of the new/improved XtxtUML features in action:**

- Uniqueness validation of class-like entities is case insensitive.
case-insensitive-uniqueness-validation
cross-file-case-insensitive-duplication

- Fully qualified names are used in type mismatch error messages if the simple names of mismatching types are equal.
fqn-in-type-mismatch-messages

- Java reserved words are no longer allowed as identifiers.
java-reserved-words

- Looser grammar, stricter validation of package and model declarations.
looser-grammar-stricter-validation

- Model declarations could be placed only in *model-info* files even in the previous implementation. Now, it is also validated that *model-info* files must contain only a model declaration.
model-info-improvements

- Multiplicity of association ends is now checked.
multiplicity-validation

- Package and model declarations are now validated.
package-and-model-declaration

- Inheritance is now considered in class property accessibility validation.
property-inheritance

- Uniqueness validation is extended to receptions.
uniqueness-of-receptions

<hr>

This pull request:
- resolves #29,
- resolves #40,
- affects #320,
- fixes #343,
- resolves #387,
- resolves #388,
- resolves #390,
- and fixes #426.

It should be noted that technically, the XtxtUML runtime is no longer dependent on the UI, therefore #320 would be resolvable here. However, I would like to create a separate pull request for that, as it is more releng-related.

### DIFF
--- a/dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/JtxtUMLCompilationParticipant.java
+++ b/dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/JtxtUMLCompilationParticipant.java
@@ -1,6 +1,11 @@
 package hu.elte.txtuml.validation;
 
+import java.io.File;
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
 
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
@@ -22,8 +27,11 @@ import hu.elte.txtuml.validation.visitors.ModelVisitor;
  * reconcile events (not typing for a few seconds) and build events
  */
 public class JtxtUMLCompilationParticipant extends org.eclipse.jdt.core.compiler.CompilationParticipant {
+
 	private static final String TXTUML_NATURE_ID = "hu.elte.txtuml.project.txtumlprojectNature"; //$NON-NLS-1$
 	public static final String JTXTUML_MARKER_TYPE = "hu.elte.txtuml.validation.jtxtumlmarker"; //$NON-NLS-1$
+
+	private Map<IJavaProject, Set<BuildContext>> javaProjectToFileSet = new HashMap<>();
 
 	@Override
 	public boolean isActive(IJavaProject project) {
@@ -57,12 +65,36 @@ public class JtxtUMLCompilationParticipant extends org.eclipse.jdt.core.compiler
 	@Override
 	public void buildStarting(BuildContext[] files, boolean isBatch) {
 		for (BuildContext file : files) {
-			validateFile(file);
+			IProject project = file.getFile().getProject();
+			IJavaProject javaProject = JavaCore.create(project);
+
+			Set<BuildContext> projectFileSet = javaProjectToFileSet.get(javaProject);
+			if (projectFileSet == null) {
+				projectFileSet = new HashSet<>();
+				javaProjectToFileSet.put(javaProject, projectFileSet);
+			}
+
+			projectFileSet.add(file);
 		}
 	}
 
-	private void validateFile(BuildContext file) {
-		java.io.File systemFile = null;
+	@Override
+	public void buildFinished(IJavaProject javaProject) {
+		Set<BuildContext> projectFileSet = javaProjectToFileSet.get(javaProject);
+		if (projectFileSet == null) { // buildFinished might be called without
+										// buildStarting
+			return;
+		}
+
+		for (BuildContext file : projectFileSet) {
+			validateFile(file, javaProject);
+		}
+
+		javaProjectToFileSet.remove(javaProject);
+	}
+
+	private void validateFile(BuildContext file, IJavaProject javaProject) {
+		File systemFile = null;
 		IPath location = file.getFile().getLocation();
 		if (location != null) {
 			systemFile = location.toFile();
@@ -71,8 +103,6 @@ public class JtxtUMLCompilationParticipant extends org.eclipse.jdt.core.compiler
 			return;
 		}
 
-		IProject project = file.getFile().getProject();
-		IJavaProject javaProject = JavaCore.create(project);
 		CompilationUnit unit = null;
 		try {
 			unit = SharedUtils.parseJavaSource(systemFile, javaProject);
--- a/dev/plugins/hu.elte.txtuml.xtxtuml.ui/META-INF/MANIFEST.MF
+++ b/dev/plugins/hu.elte.txtuml.xtxtuml.ui/META-INF/MANIFEST.MF
@@ -20,7 +20,8 @@ Require-Bundle: hu.elte.txtuml.xtxtuml;bundle-version="0.5.0";visibility:=reexpo
  org.eclipse.compare,
  hu.elte.txtuml.api.model;bundle-version="0.5.0",
  org.eclipse.jdt.core;bundle-version="3.11.1",
- hu.elte.txtuml.utils;bundle-version="0.5.0"
+ hu.elte.txtuml.utils;bundle-version="0.5.0",
+ hu.elte.txtuml.validation;bundle-version="0.5.0"
 Import-Package: org.apache.log4j
 Bundle-RequiredExecutionEnvironment: JavaSE-1.8
 Export-Package: hu.elte.txtuml.xtxtuml.ui.quickfix,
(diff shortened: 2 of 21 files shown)


nagyandras95:
I have reviewed the commits and tried some validation. It seems to OK but I have found an interesting issue:

The int is accepted as state identifier instead of it is contained by your reserved word list.
int_error

djnemeth:
@nagyandras95, I could not reproduce what you've found. Please make sure that you are testing this version of txtUML. Could you provide a step-by-step guide to reproduce the issue?

nagyandras95:
I were testing this version because the other validations worked but I could not reproduce this issue on another computer. The issue has turned up after I had tried to rewrite a valid identifier.
For example:
First I have typed: 
`state alma; ` then I rewrote `alma` to `int` and an error message was shown.

But now I am not able to reproduce neither so It must have been nothing.

djnemeth:
As I believe I have found a complete solution for #388, I'll temporarily close this pull request until I'm done with its implementation.

djnemeth:
In the meantime, @nboldi, please review cd1eafc.

djnemeth:
I believe I have managed to implement a complete solution for #388, I have updated the description of this pull request accordingly. @nagyandras95, could you please review the changes?

djnemeth:
Jenkins, retest this, please.

nagyandras95:
I have also reviewed the last commit. It seems to correct for me.

PR #433, opened 2016-09-18
Sequence diagrams

zoltanturi89:
Sequence Diagrams basic functionality implementation(with plantUML based export). 
Currently Supported:
1.  Defining Lifelines
2.  Message sending
3.  OPT,ALT,LOOP,STRICT fragments( and partial SEQ support) - using JAVA programming structures

Both the export and the model implementation has a few base unit tests defined.

Contains minor changes to the Model API( Mostly concerning signal sending - and the ability to determine the sender) and minor changes to the old export wizards to use the new plantUML based export of sequence diagrams.


nagyandras95:
retest this please

nagyandras95:
@zoltanturi89 you should set the visibility of Elte-Soft membership as public because the Jenkins cannot see that you are member of it and the tests will not be run at all.

zoltanturi89:
retest this please

nagyandras95:
retest this please

zoltanturi89:
retest this please

djnemeth:
Jenkins, retest this, please.

dev/plugins/hu.elte.txtuml.export.plantuml/src/hu/elte/txtuml/export/plantuml/generator/PlantUmlCompiler.java
```diff
@@ -0,0 +1,314 @@
...
+	 * @param lifeline
+	 *            lifeline to activate
+	 */
+	public void activateLifeline(String lifeline) {
+
+		if (!lifelineIsActive(lifeline)) {
+			compiledOutput += "activate " + lifeline + "\n";
+		}
+
+		activeLifelines.add(lifeline);
```
jtamas96: Shouldn't this be in the if statement?

dev/plugins/hu.elte.txtuml.export.plantuml/src/hu/elte/txtuml/export/plantuml/generator/PlantUmlCompiler.java
```diff
@@ -0,0 +1,314 @@
...
+	 * deactivates lifeline and does the needed supplementary operations
+	 * 
+	 * @param lifeline
+	 *            lifeline to deactivate
+	 */
+	public void deactivateLifeline(String lifeline) {
+		if (lifelineIsActive(lifeline)) {
+			activeLifelines.remove(lifeline);
+
+			if (!lifelineIsActive(lifeline)) {
```
jtamas96: I think this condition is always true because the lifeline is removed from the activeLifelines before the statement.

dev/plugins/hu.elte.txtuml.export.plantuml/src/hu/elte/txtuml/export/plantuml/generator/PlantUmlGenerator.java
```diff
@@ -0,0 +1,103 @@
...
+				messages += "\n" + ex.getMessage();
+			}
+			throw new PreCompilationError(messages);
+		}
+
+		Type superClass = preCompiler.getSuperClass();
+		while (superClass != null) {
+			CompilationUnit cu = null;
+
+			if (superClass != null) {
```
jtamas96: Is this if statement needed, since we check the same condition in the loop condition?

dev/plugins/hu.elte.txtuml.export.plantuml/src/hu/elte/txtuml/export/plantuml/generator/PlantUmlCompiler.java
kmozsi: You are right, it should not activate an already activated lifeline.
By the way this solution works now, because this method is called only from one place, and there is a check performed for this case.

dev/plugins/hu.elte.txtuml.export.plantuml/src/hu/elte/txtuml/export/plantuml/generator/PlantUmlCompiler.java
kmozsi: With the previous fix, the condition is always true indeed.

dev/plugins/hu.elte.txtuml.export.plantuml/src/hu/elte/txtuml/export/plantuml/generator/PlantUmlGenerator.java
kmozsi: It is not needed.

PR #446, opened 2016-10-18
Enable trace logging in producer_consumer demo

djnemeth:
As there's no action logging in *producer_consumer*, trace logging should be enabled to make its execution observable by the user.


djnemeth:
@nagyandras95, please review.

Issue #334, opened 2016-04-20
Implement signal inheritance in XtxtUML


kovacsgabor:

djnemeth:
Signal inheritance in XtxtUML hasn't been implemented yet. As we are generating their constructors from their attributes, in case of inheritance we would need to ensure that each base signal gets initialized *before* its derived signals.

I thought this is not achievable without heavy modifications the two-phase inference process we currently use (see #313), but now I believe this might be solvable after all.

Implementing signal inheritance would also require to ensure that type computation of the signal access expression (`sigdata`) is correct.

kovacsgabor:
I see that this is much more complicated than how much it would help so I set this as a low priority issue.

PR #448, opened 2016-11-09
Data types

### LINKED ISSUE #191: Support for data types in JtxtUML
Data types are restricted versions of classes: cannot have state machine, it is not possible to start their behavior, cannot accept messages etc.
In contrast to normal classes, data types can be used as attribute types in signals, classes and also in other data types.
They are passed and sent always as values, using copying semantics. Instances have no "identity".

We have to think (and read the UML standard) about associations, compositions for data types. Alternatively, we could introduce attribute multiplicities.

### LINKED ISSUE #192: Support for data types in XtxtUML
Based on #191.

kovacsgabor:
Resolves #191.
Also tries to resolve #192 and #334, but I am almost sure that the X part still lacks some features.


kovacsgabor:
I'll split and rework this pull request.

PR #451, opened 2016-11-16
Removed the opening of the UML model.

nboldi:


kovacsgabor:
What was the reason of this pull request? Is it still relevant?

djnemeth:
This pull request is about removing the functionality which opens the EMF-UML2 representation of exported UML models at the end of the export process. Basically, each time a model gets exported (e.g. during visualization), we automatically open the resulting XML file in the workspace of the user.

The team has agreed to drop this feature, hence this pull request. It is still relevant, I believe.

nboldi:
To drop the feature you should merge the pull request, if you find it satisfactory.

Issue #438, opened 2016-10-01
Filter the list of packages and types provided by txtUML wizards


djnemeth:
Our _[Generate C++ code from txtUML](https://github.com/ELTE-Soft/txtUML/blob/d9031bf3003ba358f78a3d18ceaa5cba9bbe56e3/dev/plugins/hu.elte.txtuml.export.cpp/src/hu/elte/txtuml/export/cpp/wizardz/TxtUMLToCppPage.java)_ and _[Generate diagrams from txtUML](https://github.com/ELTE-Soft/txtUML/blob/d9031bf3003ba358f78a3d18ceaa5cba9bbe56e3/dev/plugins/hu.elte.txtuml.export.papyrus/src/hu/elte/txtuml/export/papyrus/wizardz/VisualizeTxtUMLPage.java)_ wizards allow users to browse the project and model package they wish to use as a source, after which they can also select deployment configurations and diagram descriptions, respectively.

It's a bit inconvenient that the "model" browser lists all packages within the selected project and "deployment configuration" / "diagram description" browsers are just type selection dialogs showing all classes and interfaces available in the workspace.

It would be neat if
- model browsers listed only model packages defined in the selected project, 
- deployment configuration browsers showed only configurations and
- diagram description browsers listed only diagram descriptions.

Please note that in the last two cases, displaying types which were defined in the selected project only would be too strict. For example, I believe it's perfectly acceptable to isolate diagram descriptions from a complex model in a separate project.

djnemeth:
As suggested in #180, it might be more user-friendly to provide actual *browser* dialogs (i.e. something that *lists* the available options) for diagram descriptions and deployment configurations, instead of the currently implemented *search* dialogs.

My concerns are, however, that the list of available descriptions/configurations could grow too large in case of complex workspaces, as we cannot easily tell which model they belong to (apart from state machine diagrams). This, of course, holds only if we look for said definitions *outside* the project of the source model as well, which I think we should, see the last paragraph of my previous comment.

@dobreffandras and @kmozsi, what do you think?

dobreffandras:
Yes, you're right so for selecting _deployment configurations_ / _diagram descriptions_ the [FilteredTypesSelectionDialog](http://grepcode.com/file/repository.grepcode.com/java/eclipse.org/3.6.2/org.eclipse.jdt/ui/3.6.2/org/eclipse/jdt/internal/ui/dialogs/FilteredTypesSelectionDialog.java) is not the way to go.

[...]

PR #452, opened 2016-11-22
Diagram and cpp generation wizard improvements

kmozsi:
Features included:
- Packages are filtered in the model browser dialogs, only model packages (and subpackages) are visible.
- Diagram description and c++ configuration selection is also filtered.
- Allow to choose multiple diagram descriptions to visualize.
- Diagram selection is directly in the wizard instead of a new dialog - tree selection view.
- Allow to select diagram layouts and c++ configurations from different java projects (than the model package).


dobreffandras:
Seems to be working fine, but:

- Is it intended that the subpackages of the model are also listed? If I'm not mistaken @kovacsgabor  said that it took him effort to implement this feature, so eliminating it is theoritically not a big issue.

- I've found another issue concerning the checkboxes. If the window is reopened and all items under a package is checked the the package is not checked but marked with a box. See the picture.

issue

- The checkmarks are looking weird by the way, but I think that is not our fault, am I right?

kovacsgabor:
> Is it intended that the subpackages of the model are also listed? If I'm not mistaken @kovacsgabor said that it took him effort to implement this feature, so eliminating it is theoritically not a big issue.

As Java in fact does _not_ know the term "subpackage", we do this:

1. Search for txtUML models, that is, Java packages that have a `package-info.java` file in which the package is annotated with the `hu.elte.txtuml.api.model.Model` annotation.
2. Search for all subpackages of txtUML models, that is, Java packages whose name is like `x.y` where `x` is the name of a txtUML model.

So yes, retrieving _only_ the models themselves should be easy by performing only the first step.

dobreffandras:
The two fixes are fine. 

Somehow the checkmarks are looking good for me as well right now:
checkbox

The diagram description windows are reviewed and able to merge. @nagyandras95 please take a look at the cpp generation wizard.

kmozsi:
@dobreffandras, thanks for the review. 
The last two commits resolve these issues except the last one, which might be platform-specific, because I haven't encountered this problem. As you said, it is not our fault, as checkmarks are drawn by the Tree widget of CheckboxTreeViewer using OS calls.

djnemeth:
Did this resolve #438?

dobreffandras:
I think yes, it did. For the diagram descriptions at least:
- only model packages are listed in the model selection dialog
- only diagram descriptions (grouped by projects) are listed where you want to select the diagram descriptions.

@nagyandras95 what about the configurations?

dobreffandras:
I've found a bug. When the user clicks the browse button for the model selection the UI behavior gets corrupted in the case when the project name is empty. It's because the project searching throws an unchecked exception which wasn't handled. 

I fixed this bug. At the same time a took a look at the Cpp wizard. This issue resolves #438.

kmozsi:
> Did this resolve #438?

Currently I am working on the 'visualize' context menu, which is mentioned in #438. 
This pull request does not contain that feature, so it resolves the issue only partially. 

@dobreffandras, thanks for the bugfix.

dobreffandras:
I created a new issue about the mentioned context menu

nagyandras95:
The cpp wizard is listing only the model packages and the configuration classes, so this part seems to OK for me.

Issue #424, opened 2016-08-03
txtUML wizards should ask if modified files shall be saved


djnemeth:
Currently, txtUML wizards don't notify the user if there are modified files in the selected model, or the selected deployment configuration / diagram definition contains unsaved changes. As this can be inconvenient or even confusing, the wizards should ask the user whether the affected files shall be saved first.

PR #456, opened 2016-11-29
Save before generate

lkopi:
This pull request is intended to resolve [#424](https://github.com/ELTE-Soft/txtUML/issues/424).

If there are unsaved changes that are affected by the C++ or the diagram generation, then a dialog pops up (shown below), where the user can select the resources, that he/she wishes to save prior to continuing the generation. If this dialog is cancelled, cancels the generation.

save_and_launch

If the "Save automatically before build" option is turned on, it will automatically save all modified resources, which are affected by the generation. (This option can be found on the General > Workspace preference page.)


djnemeth:
@lkopi, please provide a short description in the first comment.

kovacsgabor:
I checked it in a fresh Eclipse installation and it works properly but it lacks one important feature: It does not check ```.xtxtuml``` files, only ```.java``` sources. Would it be hard to implement that too?

kovacsgabor:
I opened a new issue for the case of XtxtUML files (#478) to close this pull request.

PR #464, opened 2016-12-03
Diagram visualization context menu

### LINKED ISSUE #463: Diagram visualization context menu
A context menu is needed to the package/project explorer: if the user clicks on a diagram description with right button, and chooses 'Visualize', the Visualization wizard should appear with this selected description.

kmozsi:
This pull request intends to resolve #463.

- A 'Visualize diagram' option is added to the context menu of diagram descriptions in the project explorer.
- This option is only accessible in case of descriptions, otherwise it is not visible.
- Multiple selection is allowed.
- Choosing the visualize option the visualization wizard should appear, and the selected descriptions should be checked in the diagram selection tree.

### DIFF
--- a/dev/plugins/hu.elte.txtuml.export.papyrus/META-INF/MANIFEST.MF
+++ b/dev/plugins/hu.elte.txtuml.export.papyrus/META-INF/MANIFEST.MF
@@ -35,7 +35,8 @@ Require-Bundle:
  org.eclipse.emf.ecore,
  org.eclipse.jdt.ui,
  com.google.guava,
- hu.elte.txtuml.utils.eclipse
+ hu.elte.txtuml.utils.eclipse,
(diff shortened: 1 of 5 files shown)


kmozsi:

dobreffandras:
It is a strange behavior, when I select a Description (or more) the Wizard opens, but the project and model does not change accordingly. It's clear that it cannot be done, because the exact project and model cannot be determined from the selected diagrams (at least for now). But it can confuse the user very easily. 

Previously we had the same problem and decied to list all descriptions of the workspace (grouped by projects) in the visualization Wizard, even the ones that had nothing to do with the selected project-model combination. 

Considering this is the second time we came up with a solution which isn't user-friendly I propose to find a solution.

Anyway I approve this pull-request despite the limitations.

PR #474, opened 2017-01-29
Signals are allowed as operation and constructor parameters

kovacsgabor:
I think that it is a sensible update.


devaigergely81:
How do this affect the UML2 export?

kovacsgabor:
I have to admit I don't know. However, as a txtUML user I would like to perform operations on signals without having to box and unbox them all the time.

kovacsgabor:
We looked into the problem and found that in UML2, signals are allowed as operation and constructor parameters therefore this should be implementable in the model export.

dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/messages.properties
```diff
@@ -6,8 +6,8 @@ InvalidModifier_message=Invalid modifier. Only visibility modifiers are allowed
 InvalidSignalContent_message=Invalid element in signal. Only attributes and constructors are allowed.
 InvalidTemplate_message=Templates are not yet supported in txtUML models.
 InvalidTypeInModel_message=Invalid type declaration in model. Only model classes, model enums, signals, associations, interfaces and connectors can be part of the model.
-InvalidTypeWithClassAllowed_message=Invalid type. Only boolean, double, int, String, model enums and model class types are allowed.
-InvalidTypeWithClassNotAllowed_message=Invalid type. Only boolean, double, int, String, model enums and external interfaces are allowed.
+InvalidParameterType_message=Invalid type. Only boolean, double, int, String, model enums, signals and model class types are allowed.
```
dobreffandras: Missing _external interfaces_. 
From the [implementation](https://github.com/ELTE-Soft/txtUML/blob/912ac6ea4143913e218c44311f0ee656766fa5bf/dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/visitors/Utils.java#L51) it seems a parameter can be all of types that an attribute can be. Attributes can be external interfaces, but are not listed here by the parameters. (On the xtxtuml side it is correct)

dobreffandras:
As @kovacsgabor mentioned this PR is included in another(#485). So I close it, and copy my review to that.

PR #482, opened 2017-03-05
XtxtUML execution element

### LINKED ISSUE #430: Rework the 'execution' element of XtxtUML
As #331 has changed the model execution interface significantly, XtxtUML executions in their current state cannot be run. We have to adapt this element to make it directly executable while keeping its syntax and semantics as simple as possible. The main idea is that XtxtUML executions should provide a simple way to populate the given model, i.e. to instantiate objects, associations and connectors. Note that this happens _inside_ the model, therefore there's no need to provide language elements neither for [API](https://github.com/ELTE-Soft/txtUML/blob/13e6e86a70868491505f3b41b795c34ae28b46ac/dev/plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/API.java) calls nor for the new [model execution interface](https://github.com/ELTE-Soft/txtUML/blob/13e6e86a70868491505f3b41b795c34ae28b46ac/dev/plugins/hu.elte.txtuml.api.model.execution/src/hu/elte/txtuml/api/model/execution/ModelExecutor.java).

We have agreed on the following solution:
The syntax of executions doesn't need to be changed at all. Instead, only the translation of executions should be modified such that from each execution element a new Java class is generated with the same name and with two public static void functions. One of these functions should be called `init`, and it should contain the JtxtUML action code translated from the original execution body. The other function should be a conventional Java main function where a model executor should be created, initialized with the aforementioned `init` function and eventually launched with the _most typical_ configuration.

This way, executions could
- provide a simple and fast way to test models,
- be executed directly,
- be customized by creating a new Java class with a main function, where the user could define an individually configured model executor while still being able to use the generated `init` function as initializer, and perform [API](https://github.com/ELTE-Soft/txtUML/blob/13e6e86a70868491505f3b41b795c34ae28b46ac/dev/plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/API.java) calls to communicate safely with the model from the outside world.

kmozsi:
Resolves #430.

### DIFF
--- a/dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/jvmmodel/XtxtUMLJvmModelInferrer.xtend
+++ b/dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/jvmmodel/XtxtUMLJvmModelInferrer.xtend
@@ -88,22 +88,19 @@ class XtxtUMLJvmModelInferrer extends AbstractModelInferrer {
 			visibility = JvmVisibility.PUBLIC
 
 			members += exec.toMethod("init", Void.TYPE.typeRef) [
-				documentation = exec.documentation
 				visibility = JvmVisibility.PUBLIC
-
 				static = true
 				body = exec.body
 			]
 
 			members += exec.toMethod("main", Void.TYPE.typeRef) [
-				documentation = exec.documentation
 				visibility = JvmVisibility.PUBLIC
 				parameters += exec.toParameter("args", String.typeRef.addArrayTypeDimension)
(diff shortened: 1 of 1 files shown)


kmozsi:
Briefly, I chose ModelExecutor.start(Runnable), because that fits the best to our intentions: e.g. ModelExecutor.run() awaits termination, but it is not needed now. We do not want to call shutdown() either. ModelExecutor.start(Runnable) does not await termination, it returns instantly, without waiting for initialization or termination.

djnemeth:
The team has decided that the expressive power of XtxtUML executions has to be increased. The details haven't been discussed yet, therefore I close this pull request for now.

PR #483, opened 2017-03-09
Glue class fixed in garage model

kovacsgabor:
`launch` must be used in this class instead of `start`.
Read http://txtuml.inf.elte.hu/wiki/doku.php?id=v051:userguide#executing_models


kovacsgabor:
The team agreed that the best solution for this is to split up the `Glue` class into two classes which will be responsible for the two directions of communication between the UI and the model.

PR #496, opened 2017-03-16
This solves issue 450:

sallyhalal:
When a Multiple or HiddenMultiple multiplicity without Min and / or Max
annotation is exported to uml2 model, the multiplicity is 1..1 instead
of 0..* as documented in the Language Guide.

the handeling was written but not referenced in MultiplicituProvider
class so I fixed that by adding the function isMultiple.


kovacsgabor:
ok to test

PR #502, opened 2017-03-26
Fix save before generate

### LINKED ISSUE #501: Cannot visualize when there are unsaved changes in an already generated diagram
The visualization (and c++ gen) wizard throws NullPointerException and does not generate anything when there are unsaved changes in an already generated diagram.

kmozsi:
Resolves #501.


djnemeth:
Using a runtime Eclipse I have reproduced #501 on `master`, and then verified that this fix indeed resolves it. Looks good to me. 👍

Issue #321, opened 2016-04-11
J- and XtxtUML elements are not fully interchangeable


djnemeth:
Elements (e.g. classes) defined in JtxtUML can't be referenced in all structural parts of an XtxtUML model. For example, a class written in JtxtUML can't be specified as the type of an association end in XtxtUML. This is due to the Xtext feature discussed in #226.

J- and XtxtUML elements should be fully interchangeable, just like Java and Xtend classes are. While this modification heavily affects the overall implementation of XtxtUML, it is also inevitable: classes with external components (see #254) defined in Java must be accessed through associations in the model. In the current XtxtUML implementation however,  this is not possible.

devaigergely81:
Proposal for short term: In an X-syntax model the classes with external elements will also be X-classes, but their external attributes can be typed by simple Java classes (?) and the activities in the X class can reference and use the Java classes.

kovacsgabor:
I found a solution for this problem.

We can introduce a new language feature in X to declare a class which is defined in J.

The users will have to write:
```
jtxtuml-class A;
```
to use the class `A` in X.

I am currently working on this and it seems to work properly.

kovacsgabor:
`Note:` the keyword for the above mentioned language feature is up for discussion. It is also possible in this case to split it into two separate words: `jtxtuml class`

kovacsgabor:
My proposed solution does not solve the general interchangeability of J and X elements but enables the use of model classes defined in J to be used in X which was the original problem that caused the opening of this issue (as @djnemeth wrote in the original comment).

Also, this is much easier to implement then the general interchangeability which would require a rewrite of the whole XtxtUML implementation.

kovacsgabor:
I see that it was not documented here but the team previously agreed to stay with the current XtxtUML implementation and do not solve the general interchangeability as it would require much more effort than it is worth. My proposal, however, is much easier to implement and we would gain a lot from it, especially in the case of external classes.

kovacsgabor:
Further testing in the matter proved that my proposed solution does not work as fine as it seemed. (New features in an Xtext DSL easily become extremely complex from one step to another.)

In the future, it might be needed to support full interchangeability by reworking the XtxtUML implementation.

Issue #258, opened 2016-02-25
Static attributes and references


kovacsgabor:
As stated in #61, static attributes are problematic when it comes to parallelism. However, they raise another question, namely the problem of references. Can a static attribute be a reference to a model object (or a collection of references)?

kovacsgabor:
@ELTE-Soft/all, what do you think?

Issue #196, opened 2015-12-10
Static is not allowed in XtxtUML


devaigergely81:
Static operations and attributes are allowed in JtxtUML, but cause parse error in XtxtUML.

djnemeth:
According to #61, it still has to be ensured that the use of Java keyword `static` is compatible with the JtxtUML runtime. This is the main reason why it's not implemented in XtxtUML.

Do you think we should allow the keyword in the grammar but raise a validation error on its occurence until #61 is resolved?

Issue #61, opened 2015-10-22
Implementation of static elements


kovacsgabor:
In the current state of the JtxtUML API, static classes, attributes and methods can be defined using the Java keyword ```static```. However, further investigaton is needed to ensure its compatibility with other components of the API, as well as with other modules of the txtUML project.

kovacsgabor:
I'll split the problem into three minor ones: the problem of static classes, attributes and methods.

Static (nested) classes are not a problem anymore since our models are split into multiple files, as model classes, associations and signals are no more nested classes. We still have nested classes, that's true (like states, transitions, ports, association ends, etc.), but they must be non-static for other reasons so it is a problem of validation from this point.

Static methods should be problematic in no way.

Static attributes should also work, unless:
 * JtxtUML models become multi-threaded. This is a problem, however, which we are well aware of (see #145).
 * The user starts multiple executors in parallel which run the same model. We must either ensure that this may never happen or warn the users in the documentations.

kovacsgabor:
Static methods should be supported.

Whether we support static attributes or not will depend on our discussions about parallel models.

PR #525, opened 2017-07-01
New syntax for externals in both J and X

### LINKED ISSUE #14: External class support in XtxtUML?
Is there support for defining external classes in the custom syntax?

### LINKED ISSUE #254: External class support in JtxtUML

### LINKED ISSUE #499: Simplify communication between the model and outside world

kovacsgabor:
Resolves #254.
Resolves #14.
Introduces the `static` keyword in X (only allowed on operations and external attributes at the moment). See #61, #196 and #258 for discussions about static elements.

[UPDATE Oct 27 2017] Resolves #499 with the introduction of the World stdlib model.

The syntax proposed in #254 is used in J. In X, the keywords `external` and `external-body` can be used.

Currently supported:
 * external attributes, operations and constructors in both J and X,
 * external member classes in J,
 * fully external classes in J (that is, when the full class is marked external),
 * external interfaces in J. When an external class implements an interface which is not part of the model, it can mark that interface (after the `implements` keyword). Note that as currently we do not have support for classes that implement modeled interfaces, all interfaces that are implemented by a model class are ignored. It is not even validated whether an interface implemented by a class is part of the model or not. Therefore, using this feature is optional at the moment but I recommend it for the future.

Limitations:
 * The standard library (`hu.elte.txtuml.api.stdlib` project) is no longer auto-imported in a txtUML project. It has to be manually added to the classpath and be imported to the workspace. (In the new txtUML Project wizard there is a new option which lets the user add the stdlib project to the classpath.)
 * As external classes are now simple model classes, they must be connected to other classes via associations (they cannot be stored in an attribute). Because of this, if an instance of an external class has to be stored, it can only be used by a class written in X if the external class is also written in X. See #321 for details. 

Validation, UML2 export and C++ generation are also updated.

*Note:* after this update, it will soon become very confusing if we keep on calling these classes "external classes" as they are now "classes with external features" opposed to those classes I called in this post "fully external classes" which are in package of a model but are marked external as a whole and therefore are completely invisible from the model. They should be called "external classes" because in case of attributes and operations, we also call external those which are completely invisible from the model.


djnemeth:
@nagyandras95, please review the parts related to UML2 and C++ export.

dev/plugins/hu.elte.txtuml.xtxtuml.ui/src/hu/elte/txtuml/xtxtuml/ui/contentassist/XtxtUMLReferenceProposalCreator.xtend
```diff
@@ -274,16 +271,12 @@ class XtxtUMLReferenceProposalCreator extends XbaseReferenceProposalCreator {
 			// be difficult to achieve starting from a plain JvmType
 			proposedObj instanceof JvmDeclaredType && (proposedObj as JvmDeclaredType).superTypes.exists [
 				val typeRef = toLightweightTypeReference;
-				typeRef.isSubtypeOf(DataType) || typeRef.isInterface && typeRef.isSubtypeOf(ExternalType) || typeRef.isSubtypeOf(ModelEnum) ||
```
djnemeth: In case of the enclosing entity is marked as external, we should not filter type proposals.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/XtxtUML.xtext
```diff
@@ -212,10 +212,18 @@ TUMultiplicity :
 
 // Commons
 
+TUModifiers:
```
djnemeth: Isn't this too complicated (and too strict, with regards to the order)? See rules [`CommonModifier`](https://github.com/eclipse/xtext-xtend/blob/314ca2fd50b88f8c585dd643c2234008a1148327/org.eclipse.xtend.core/src/org/eclipse/xtend/core/Xtend.xtext#L223-L226) and [`XtendTypeDeclaration`](https://github.com/eclipse/xtext-xtend/blob/314ca2fd50b88f8c585dd643c2234008a1148327/org.eclipse.xtend.core/src/org/eclipse/xtend/core/Xtend.xtext#L26) from the Xtend grammar.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/validation/XtxtUMLExternalityHelper.xtend
```diff
@@ -0,0 +1,63 @@
...
+import java.util.HashMap
+import java.util.Map
+import org.eclipse.emf.ecore.EObject
+import org.eclipse.xtext.xbase.XBlockExpression
+
+import static hu.elte.txtuml.xtxtuml.xtxtUML.TUExternality.*
+
+class XtxtUMLExternalityHelper {
+
+	val Map<EObject, TUExternality> externality = new HashMap
```
djnemeth: I believe this is not populated in `calculateExternality`.

nagyandras95:
One commit is missing from cpp-externals, but I will get it from @petra95

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/formatting2/XtxtUMLFormatter.xtend
```diff
@@ -137,8 +138,14 @@ class XtxtUMLFormatter extends XbaseFormatter {
...
 	}
 
+	def dispatch void format(TUModifiers it, extension IFormattableDocument document) {
+		regionFor.feature(TU_MODIFIERS__STATIC).append[oneSpace];
+		regionFor.feature(TU_MODIFIERS__EXTERNALITY).append[oneSpace];
+		regionFor.feature(TU_MODIFIERS__VISIBILITY).append[oneSpace];	
+	}
+	
 	def dispatch void format(TUConstructor it, extension IFormattableDocument document) {
-		regionFor.feature(TU_CONSTRUCTOR__VISIBILITY).append[oneSpace];
```
djnemeth: I recommend to prepend the `name` feature with one space here. Of course this rule will be redundant (see the previous comment), but I believe this is more in line with the aforementioned invariant.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/formatting2/XtxtUMLFormatter.xtend
```diff
@@ -150,7 +157,7 @@ class XtxtUMLFormatter extends XbaseFormatter {
 	}
 
 	def dispatch void format(TUAttributeOrOperationDeclarationPrefix it, extension IFormattableDocument document) {
-		regionFor.feature(TU_ATTRIBUTE_OR_OPERATION_DECLARATION_PREFIX__VISIBILITY).append[oneSpace];
```
djnemeth: I recommend to prepend the `type` feature with one space here. Of course this rule will be redundant (see the comment before the previous one), but I believe this is more in line with the aforementioned invariant.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/validation/XtxtUMLModifierValidator.xtend
```diff
@@ -0,0 +1,38 @@
...
+
+import hu.elte.txtuml.xtxtuml.xtxtUML.TUConstructor
+import org.eclipse.xtext.validation.Check
+
+import static hu.elte.txtuml.xtxtuml.validation.XtxtUMLIssueCodes.*
+import static hu.elte.txtuml.xtxtuml.xtxtUML.XtxtUMLPackage.Literals.*
+import hu.elte.txtuml.xtxtuml.xtxtUML.TUAttribute
+import hu.elte.txtuml.xtxtuml.xtxtUML.TUExternality
+
+class XtxtUMLModifierValidator extends XtxtUMLConnectorValidator {
```
djnemeth: As constructors and attributes are named elements, it's advisable to include their names in the error messages, because in this way it's easier for the user to manage issues in the problems view. For an example, see _[XtxtUMLClassValidator.xtend:79](https://github.com/ELTE-Soft/txtUML/blob/5ea692b4cb20a19029876b3edccc1db6d916d78e/dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/validation/XtxtUMLClassValidator.xtend#L79-L80)_.

dev/tests/hu.elte.txtuml.xtxtuml.tests/src/hu/elte/txtuml/xtxtuml/tests/parser/XtxtUMLParserTestUtils.xtend
```diff
@@ -136,7 +136,7 @@ class XtxtUMLParserTestUtils {
 		assertTrue(member instanceof TUAttribute);
 		val attribute = member as TUAttribute;
 
-		assertEquals(visibility, attribute.prefix.visibility);
+		assertEquals(visibility, attribute.prefix.modifiers.visibility);
```
djnemeth: Were the tests for the new elements skipped intentionally? I understand that our primary concern here is to check the viability of the concept of externals, but I'm not entirely comfortable with omitting tests.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/timers/Timer.xtxtuml
```diff
@@ -28,18 +35,38 @@
 	 *            the signal which is to be sent after the delay
 	 * @param millisecs
 	 *            the time in millisecs to wait before sending the signal
-	 * @return a timer instance to manage this delayed send operation before it
+	 * @return a handle object to manage this delayed send operation before it
 	 *         happens
 	 */
-	static Timer start(ModelClass targetObj, Signal signal, int millisecs) {
-		return new TimerImpl(targetObj, signal, millisecs);
+	public static external-body Timer start(Signal sig, ModelClass targetObj, int millisecs) {
```
djnemeth: Is `external-body` necessary here? The called `create` seems to be an `external-body` function as well. The same goes for [`add`](https://github.com/ELTE-Soft/txtUML/blob/5ea692b4cb20a19029876b3edccc1db6d916d78e/dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/timers/Timer.xtxtuml#L87-L99).

examples/demo/clock/src/clock/x/model/clock.xtxtuml
```diff
@@ -1,17 +1,17 @@
...
+import clock.x.model.connectors.MinuteValues;
 import clock.x.model.connectors.Minutes;
+import clock.x.model.connectors.SecondValues;
 import clock.x.model.connectors.Seconds;
-import clock.x.model.associations.PendulumInClock;
-import clock.x.model.associations.SecondHand;
-import clock.x.model.associations.MinuteHand;
-import clock.x.model.associations.HourHand;
-import clock.x.model.associations.DisplayInClock;
+import java.util.concurrent.Callable;
```
djnemeth: Unused "external" import.

dev/plugins/hu.elte.txtuml.export.cpp/src/hu/elte/txtuml/export/cpp/templates/structual/FunctionTemplates.java
```diff
@@ -13,37 +14,54 @@ public static String functionDecl(String functionName) {
...
 	public static String functionDecl(String functionName, List<String> params) {
-		return FunctionTemplates.functionDecl(ModifierNames.NoReturn, functionName, params);
+		return FunctionTemplates.functionDecl(ModifierNames.NoReturn, functionName, params, "", false);
 	}
 
 	public static String simpleFunctionDecl(String returnType, String functionName) {
 		return PrivateFunctionalTemplates.cppType(returnType) + " " + functionName + "()";
 	}
 
 	// TODO modifiers (static, etc) - not supported yet
```
kovacsgabor: @nagyandras95, what do you think?

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/XtxtUML.xtext
```diff
@@ -212,10 +212,18 @@ TUMultiplicity :
...
+	{TUModifiers} (static?='static'? & visibility=TUVisibility? & externality=TUExternality?)
```
kovacsgabor: Try to remove `{TUModifiers}`. Result: Xtext warning.

> The rule 'TUModifiers' may be consumed without object instantiation. Add an action to ensure object creation, e.g. '{TUModifiers}'.

Try to remove the parentheses. Result: Xtext error.

> Actions may not be used in unordered groups.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/formatting2/XtxtUMLFormatter.xtend
```diff
@@ -137,8 +138,14 @@ class XtxtUMLFormatter extends XbaseFormatter {
 		}
 	}
 
+	def dispatch void format(TUModifiers it, extension IFormattableDocument document) {
```
kovacsgabor: What is your proposal then? To write a comment or to use the offset to determine the order of the regions. The latter seems a bit messy for me.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/validation/XtxtUMLExternalityHelper.xtend
```diff
@@ -0,0 +1,63 @@
...
+	private def calculateExternality(EObject object) {
+		val sup = object.eContainer?.externality ?: NON_EXTERNAL
+		switch (sup) {
+			case EXTERNAL:
+				return EXTERNAL
+			case EXTERNAL_BODY:
+				return if(object instanceof XBlockExpression) EXTERNAL else NON_EXTERNAL
+			case NON_EXTERNAL: {
+				var TUModifiers modifiers
+				switch (object) {
```
kovacsgabor: Simplified.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/validation/XtxtUMLExternalityHelper.xtend
```diff
@@ -0,0 +1,63 @@
...
+		val ret = externality.get(object)
+
+		if (ret == null) {
+			object.calculateExternality
+		} else {
+			ret
+		}
+	}
+
+	private def calculateExternality(EObject object) {
```
kovacsgabor: I added JavaDoc to the class to explain everything.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/timers/Timer.xtxtuml
kovacsgabor: The `create` here is the `Action.create` API method which in the exported UML2 model is an object creation action followed by a constructor call. However, the constructor which is called here is external.

Note that the methods of the `Action` class are all treated specially when the model is exported to UML2. They are marked to have an external body only to emphasize for the users that they can be used in the model but their implementation is not modeled. It may also simplify the validation of action language elements.

Another reason to use `external-body` here and also in the case of the `add` method is that the `Timer` class has been intended to be just an interface in the model above an external feature. When exporting a model, the target language has complete freedom in how to implement this class.

dev/tests/hu.elte.txtuml.xtxtuml.tests/src/hu/elte/txtuml/xtxtuml/tests/parser/XtxtUMLParserTestUtils.xtend
kovacsgabor: Yes, the tests were skipped intentionally. If you write them, I will happily include them in this pull request.

examples/demo/clock/src/clock/x/model/clock.xtxtuml
kovacsgabor: Fixed.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/validation/XtxtUMLModifierValidator.xtend
kovacsgabor: Done.

dev/plugins/hu.elte.txtuml.xtxtuml.ui/src/hu/elte/txtuml/xtxtuml/ui/contentassist/XtxtUMLReferenceProposalCreator.xtend
kovacsgabor: The enclosing entity of the part of the code where we are about to propose or the enclosing entity of the proposal? I think the first one, but I have no idea how to implement this.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/math/Math.xtxtuml
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * Proxy class for the {@linkplain java.lang.Math} class.
+ */
+class Math {
+
+	/**
+	 * This class cannot be instantiated from the model.
+	 */
+	private external-body Math() {}
```
kovacsgabor: I wanted to emphasize the fact that these types are only interfaces above some external features in the model and therefore their constructors are present in the model only to prohibit their instantiation. The body of these constructors (if any) has no part in the model.

dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/visitors/ModelVisitor.java
```diff
@@ -18,6 +18,10 @@ public ModelVisitor(ProblemCollector collector) {
 	public boolean visit(TypeDeclaration elem) {
 		Utils.checkTemplate(collector, elem);
```
kovacsgabor: Fixed.

kovacsgabor: Fixed.

dev/plugins/hu.elte.txtuml.xtxtuml.ui/src/hu/elte/txtuml/xtxtuml/ui/contentassist/XtxtUMLReferenceProposalCreator.xtend
djnemeth: I meant your first interpretation. I think it's enough if we check the _model_ (the part of the code where we are about to create proposals) using the externality helper, but it will require some testing as I'm not sure if the model is just a proxy here, see #507. I believe this modification could be placed right before _[XtxtUMLReferenceProposalCreator.xtend:245](https://github.com/ELTE-Soft/txtUML/blob/1d5ff11b2d8a354b1a90e861ced5ef50f9c6aef9/dev/plugins/hu.elte.txtuml.xtxtuml.ui/src/hu/elte/txtuml/xtxtuml/ui/contentassist/XtxtUMLReferenceProposalCreator.xtend#L245)_.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/XtxtUML.xtext
djnemeth: I see now, it is required because all features of this rule are optional and during parsing, Xtext instantiates AST nodes only at the first actual assignment of a feature of the corresponding rule.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/XtxtUML.xtext
djnemeth: We should consider that using a common enum rule for all possible modifiers would result in more intuitive error messages. For example with this solution, in case of a duplicate modifier we can produce only a syntax error instead of an error message mentioning the duplication, where the latter would be achievable with enum rules and validation.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/formatting2/XtxtUMLFormatter.xtend
djnemeth: Actually it can be done in a two-liner, but I can see your concerns. An example implementation:
```Java
def dispatch void format(TUModifiers mods, extension IFormattableDocument document) {
    #[TU_MODIFIERS__STATIC, TU_MODIFIERS__EXTERNALITY, TU_MODIFIERS__VISIBILITY]
        .map[mods.regionFor.feature(it)].sortBy[it?.offset].tail.forEach[prepend[oneSpace]];
}
```
Experimenting with the solution above I realized that with the current grammar in fact it's not possible to keep the aforementioned invariant, as for this we would need to put a space between the `modifiers` and `type` features of declaration prefixes. This can't be done because these are so-called _containment references_ and as such, they aren't supported by the region finder we use.

So the conclusion is that if we keep the initially proposed grammar, we have no other option than writing a comment about the violation and leaving the implementation unchanged. However, we still might have to consider the enum-based solution I mentioned in my [comment](https://github.com/ELTE-Soft/txtUML/pull/525#discussion_r125604117) about the grammar, which solution would also make my second proposal feasible here, as enum features are still supported by the region finder.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/formatting2/XtxtUMLFormatter.xtend
djnemeth: > I recommend to prepend the `type` feature with one space here.

Unfortunately, this is not possible. See my previous [comment](https://github.com/ELTE-Soft/txtUML/pull/525#discussion_r126102659) for an explanation.

dev/tests/hu.elte.txtuml.xtxtuml.tests/src/hu/elte/txtuml/xtxtuml/tests/parser/XtxtUMLParserTestUtils.xtend
djnemeth: Okay, I'll write the missing tests after we resolved all questions.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/XtxtUML.xtext
kovacsgabor: First of all, using the `&` operator is the proposed solution for this problem in Xtext, despite the lack of detailed error messages (which still can be achieved, as Xtext provides the ability to write custom syntax error messages, see [this stackoverflow question](https://stackoverflow.com/questions/14526524/xtext-customizing-error-msg-by-unordered-groups) or [this Eclipse forum question](https://www.eclipse.org/forums/index.php/t/452330/)).

Secondly, in my opinion, the whole txtUML lacks too many important features to bother with such enhancements that slightly improve the user experience. If we change it to an enum rule, we have to manually write the required checks for duplications and also a code which retrieves the required modifiers from the the enum list.

Not to mention that it is not a clean solution for me to duplicate the visibility modifiers in a `TUModifier` and in a `TUVisibility` rule because at multiple places, only visibility modifiers are used in the language and there is no reason to syntactically enable writing a `static` or an externality keyword there.

kovacsgabor: Nevertheless, if you write the enum rule version, I approve it.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/formatting2/XtxtUMLFormatter.xtend
kovacsgabor: If this is not possible in case of an attribute or operation declaration prefix, I think we should not do this in this case either. Also, I do not like at all the idea to do something twice only to preserve some invariant which we cannot really preserve. As I see it, it is a lot worse than simply admitting the failure to preserve the invariant and not just because it reduces performance but because it is a lie! If someone reads the code and finds no explicit information about this violation, they simply assume that the invariant is preserved and add changes to the code that result in surprising failures.

 I think the only clean solution here is to add an explanatory comment here.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/timers/Timer.xtxtuml
djnemeth: > The `create` here is the `Action.create` API method which in the exported UML2 model is an object creation action followed by a constructor call. However, the constructor which is called here is external.

So if I understand it correctly, the `external-body` modifier is required on `start` because `Action.create` is actually unfolded in place into two actions by the model exporter, where one of these actions is external, as it calls an external constructor.

Does this imply that even though `Action.create` is labeled as `external-body` (i.e. it should be safe to call it from pure action code), the user will have to manually check the implicitly called constructor and mark the the enclosing operation with `external-body` if it's external? If yes, isn't this counter-intuitive?

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/formatting2/XtxtUMLFormatter.xtend
kovacsgabor: Possible or not, I strongly recommend an explanatory comment here instead of the added space. See my [comment](https://github.com/ELTE-Soft/txtUML/pull/525#discussion_r126123723).

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/validation/XtxtUMLExternalityHelper.xtend
kovacsgabor: Fixed in https://github.com/ELTE-Soft/txtUML/pull/525/commits/380647e8eb26d65b0722583fe3c983c2a9797a7b.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/XtxtUML.xtext
djnemeth: I agree with your points and I've totally forgot about the customizability of syntax error messages. The implementation can stay as it is now, and we should customize the error messages later.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/formatting2/XtxtUMLFormatter.xtend
djnemeth: I agree with your points.

dev/plugins/hu.elte.txtuml.xtxtuml.ui/src/hu/elte/txtuml/xtxtuml/ui/contentassist/XtxtUMLReferenceProposalCreator.xtend
djnemeth: Turns out this is not that simple, I'll take a look.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/timers/Timer.xtxtuml
djnemeth: This comment is here only not to let this question be forgotten. @kovacsgabor

dev/plugins/hu.elte.txtuml.xtxtuml.ui/src/hu/elte/txtuml/xtxtuml/ui/contentassist/XtxtUMLReferenceProposalCreator.xtend
djnemeth: Done in d6c9c47, see the full commit message and my [comment](https://github.com/ELTE-Soft/txtUML/issues/507#issuecomment-314433145) for details.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/timers/Timer.xtxtuml
devaigergely81: Results of an internal discussion (@kovacsgabor):
* It is OK that the `Timer(Signal sig, ModelClass targetObj, int millisecs)` is external, because we do not allow `Signal` and `ModelClass` as parameter types. (TODO: Check if this could be allowed later: Does UML standard allow this? Is it hard to implement in C++? @nboldi, @nagyandras95)
* We expect the user to know that an `Action.create` calls a constructor, and, if the constructor is external, then the create action can only be called in external or external-body context.

dev/plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/Action.java
```diff
@@ -50,6 +50,7 @@
 	 * @throws NullPointerException
 	 *             if <code>classType</code> is <code>null</code>
 	 */
+	@ExternalBody
```
devaigergely81: @kovacsgabor, Why do we have `@ExternalBody` here? This code is not part of a model.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/common/XtxtUMLExternalityHelper.xtend
```diff
@@ -0,0 +1,103 @@
...
+ * Querying the externality of an {@link EObject} which is not an XtxtUML
+ * language element does not yield an error, the object is simply considered
+ * non-external (as obviously neither of the other two can be applied). 
+ */
+class XtxtUMLExternalityHelper {
+
+	val Map<EObject, TUExternality> externalityMap = new HashMap
+
+	def isNonExternal(EObject object) {
+		return object.externality === NON_EXTERNAL
```
djnemeth: @kovacsgabor, is there a specific reason you used `===` here instead of `==`?

dev/plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/Action.java
djnemeth: @devaigergely81, according to the commit message of 84b1eca,

> This is to emphasize which methods of the API should be used in the model and which not.

Is this a valid reason?

dev/plugins/hu.elte.txtuml.export.cpp/src/hu/elte/txtuml/export/cpp/templates/structual/FunctionTemplates.java
kovacsgabor: As I see, this comment has already been removed.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/common/XtxtUMLExternalityHelper.xtend
kovacsgabor: It is more efficient. (And the result of the two operators must be the same as we compare enum constants here.)

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/timers/Timer.xtxtuml
kovacsgabor: We DO allow `Signal` and `ModelClass` as parameter types.

`Timer(Signal, ModelClass, int)` is external to force the user to use the static method `Timer.start(Signal, ModelClass, int)` which shows better that we are simply sending the signal to the model object like we do it in the case of the similar `Action.send(Signal, ModelClass)` method. The only difference is that in case of timers, we add a delay and get a handle object as a result which can be used to manipulate this delayed send action until it is executed.

dev/plugins/hu.elte.txtuml.api.model/src/hu/elte/txtuml/api/model/Action.java
kovacsgabor: @djnemeth answered the question for me.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/formatting2/XtxtUMLFormatter.xtend
kovacsgabor: I added the necessary comments.

dev/plugins/hu.elte.txtuml.xtxtuml/src/hu/elte/txtuml/xtxtuml/formatting2/XtxtUMLFormatter.xtend
kovacsgabor: I added the necessary comments.

dev/plugins/hu.elte.txtuml.export.cpp/src/hu/elte/txtuml/export/cpp/templates/structual/FunctionTemplates.java
djnemeth: Indeed, got ninja removed in [`722c2d4`](https://github.com/ELTE-Soft/txtUML/commit/722c2d4262d76aa8585cfdf4283dc535f885d973#diff-a22413200eda9887ce00a39479e932d1L56).

dev/tests/hu.elte.txtuml.xtxtuml.tests/src/hu/elte/txtuml/xtxtuml/tests/parser/XtxtUMLParserTestUtils.xtend
djnemeth: Done in 938ee97, 821167c, 8be57fe and a4cf200.

djnemeth:
@kovacsgabor, launching the _vending_machine_ demo results in an exception. Could you take a look?
 Here's the stack trace:
```
Exception in thread "main" hu.elte.txtuml.api.model.error.MissingRuntimeContextError: Missing runtime context. The current thread is not a RuntimeContext.
    at hu.elte.txtuml.api.model.Runtime.currentRuntime(Runtime.java:53)
    at hu.elte.txtuml.api.model.ModelClass.createRuntimeInfo(ModelClass.java:183)
    at hu.elte.txtuml.api.model.ModelClass.createRuntimeInfo(ModelClass.java:1)
    at hu.elte.txtuml.api.model.Runtime$Described.<init>(Runtime.java:160)
    at hu.elte.txtuml.api.model.StateMachine.<init>(StateMachine.java:183)
    at hu.elte.txtuml.api.model.ModelClass.<init>(ModelClass.java:177)
    at vending_machine.model.View.<init>(View.java:13)
    at vending_machine.model.View.<init>(View.java:13)
    at vending_machine.model.View$SingletonHelper.<clinit>(View.java:18)
    at vending_machine.model.View.getInstance(View.java:23)
    at vending_machine.ui.UI.<init>(UI.java:72)
    at vending_machine.ui.UI.main(UI.java:252)
```

kovacsgabor:
@djnemeth, I looked at the problem you reported in the *vending_machine* demo. The good news is I found the source of the problem. The bad news is it is a design error I made in both the *vending_machine* and the *garage* examples. I will think about a good alternative structure for these examples.

nagyandras95:
retest this please

nagyandras95:
retest this please

kovacsgabor:
retest this please

kovacsgabor:
retest this please

kovacsgabor:
@djnemeth, please review my last commits. I think that this pull request is ready to be merged.

dev/plugins/hu.elte.txtuml.api.stdlib/.gitignore
```diff
@@ -1,4 +1,3 @@
 bin/
 target/
-src-gen/*
```
djnemeth: @kovacsgabor, could you please provide some details about the compilation problem you mentioned in the commit message of e2f6d3a?

djnemeth: I believe this is caused by the very same reason we need to import the stdlib project alongside the demo models. That is, we have no Maven plugin to define when and how to compile XtxtUML files during the build process. While the ultimate solution would be to create such a plugin, I agree that versioning the files generated by the IDE is enough for now.

djnemeth:
Jenkins, retest this, please.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/world/AbstractWorldObjectListener.java
```diff
@@ -0,0 +1,25 @@
+package hu.elte.txtuml.api.stdlib.world;
+
+import hu.elte.txtuml.api.model.External;
+
+/**
+ * Abstract {@link WorldObjectListener} to define custom listeners. Override the
+ * {@link #acceptAny} method with the use of the {@link #invokeHandler} method.
+ * <p>
+ * Intended for complex use cases. Use the simpler {@code WorldObjectListener}
+ * versions is possible.
```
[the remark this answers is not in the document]
kovacsgabor: Fixed.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/world/WorldObject.xtxtuml
```diff
@@ -0,0 +1,73 @@
+package hu.elte.txtuml.api.stdlib.world;
+
+import hu.elte.txtuml.api.model.Signal;
+
+/**
+ * Represents an active entity outside the model that can accept certain signals.
+ * Belongs to a {@link World}, more precisely, to the only existing {@link World} instance in
+ * the
```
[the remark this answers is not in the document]
kovacsgabor: Fixed.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/world/WorldObjectListener.java
```diff
@@ -0,0 +1,187 @@
...
+ * <li>may have an arbitrary name.</li>
+ * </ul>
+ * <p>
+ * If multiple signal handler methods can be applied for a signal, then the one
+ * with the most specific parameter type will be invoked. If more handlers exist
+ * with the same parameter, then an unspecified one will be chosen. If no
+ * handlers can be applied, the {@link #notHandledSignal} method is called which
+ * shows a warning message by default.
+ * <p>
+ * <i>Note:</i> Instead of this class, create a subclass of its predefined
```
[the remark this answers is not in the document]
kovacsgabor: Fixed.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/world/WorldObjectListener.java
```diff
@@ -0,0 +1,187 @@
...
+ * @see AbstractWorldObjectListener
+ * @see SwingWorldObjectListener
+ * @see QueueWorldObjectListener
+ * @see ImmediateWorldObjectListener
+ */
+@External
+public abstract class WorldObjectListener {
+
+	/**
+	 * Instead of {@link WorldObjectListener}, create a subclass of its
```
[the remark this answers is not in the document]
kovacsgabor: Fixed.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/world/WorldObjectListener.java
```diff
@@ -0,0 +1,187 @@
...
+	 * Note that the registration is valid and completed even if the specific
+	 * world object does not yet exist. In this case, the world object will be
+	 * connected automatically to this listener whenever that is created.
+	 */
+	public void register(BaseModelExecutor forExecutor, String identifier) {
+		WorldHelper.getOrCreateHelperInstance(forExecutor).getOrCreateObjectHelper(identifier).setListener(this);
+	}
+
+	/**
+	 * If any listener is registered for world object that belongs to the given
```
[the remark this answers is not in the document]
kovacsgabor: Fixed.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/world/model-info.xtxtuml
```diff
@@ -0,0 +1,151 @@
...
+ * and world object identifiers are unique only in the context of a specific model execution.
+ * Multiple parallel model executions are not related in any way.
+ * 
+ * <h1>How to use</h1>
+ * 
+ * <h2>Shared between the model and the outside world</h2>
+ * 
+ * There are only two things that need to be shared between the model and the outside world.
+ * <ul>
+ * <li>The strings that identify the outside entities and their the world objects that represent them.<br/>
```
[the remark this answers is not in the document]
kovacsgabor: Fixed.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/world/model-info.xtxtuml
```diff
@@ -0,0 +1,151 @@
...
+ * {@link hu.elte.txtuml.api.stdlib.world.SignalToWorld SignalToWorld}</b>. Apart from simple Java
+ * visibility, nothing else is required for them to be used in the outside world as well.</li>
+ * </li>
+ * </ul>
+ * 
+ * <h2>Send signals from the model</h2>
+ * 
+ * Simple as it can be. Just:
+ * <ol>
+ * <li>Create and instance of the required signal (that extends {@code SignalToWorld}).</li>
```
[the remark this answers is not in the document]
kovacsgabor: Fixed.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/world/model-info.xtxtuml
```diff
@@ -0,0 +1,151 @@
...
+ * 
+ * <h2>Handle signals outside</h2>
+ * 
+ * <ol>
+ * <li>Subclass one of the subclasses of {@link 
+ * hu.elte.txtuml.api.stdlib.world.WorldObjectListener WorldObjectListener}.<br/>
+ * This is the step that has to be done with some care as different situations require different
+ * listeners to ensure thread safety. Take a look at the predefined versions or use
+ * {@link hu.elte.txtuml.api.stdlib.world.AbstractWorldObjectListener AbstractWorldObjectListener} if
+ * nothing other helps.</li>
```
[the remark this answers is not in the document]
kovacsgabor: Fixed.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/world/model-info.xtxtuml
```diff
@@ -0,0 +1,151 @@
...
+ * <li>In the listener, create signal handler methods that will take the desired signal as parameter.
+ * They also have to be {@code public} and have the
+ * {@link hu.elte.txtuml.api.stdlib.world.WorldObjectListener.SignalHandler SignalHandler} annotation.</li>
+ * <li>Call the {@link hu.elte.txtuml.api.stdlib.world.WorldObjectListener#register
+ * WorldObjectListener.register} to register for signals of the world object that has the specified
```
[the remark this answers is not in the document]
kovacsgabor: Fixed.

dev/plugins/hu.elte.txtuml.api.stdlib/src/hu/elte/txtuml/api/stdlib/world/model-info.xtxtuml
```diff
@@ -0,0 +1,151 @@
...
+ * nothing other helps.</li>
+ * <li>In the listener, create signal handler methods that will take the desired signal as parameter.
+ * They also have to be {@code public} and have the
+ * {@link hu.elte.txtuml.api.stdlib.world.WorldObjectListener.SignalHandler SignalHandler} annotation.</li>
+ * <li>Call the {@link hu.elte.txtuml.api.stdlib.world.WorldObjectListener#register
+ * WorldObjectListener.register} to register for signals of the world object that has the specified 
+ * identifier and belongs to the given model executor.</li>
+ * <li>Wait for signals to arrive.</li>
+ * </ol>
+ * <h2>Send response to the model</h2>
```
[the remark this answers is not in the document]
kovacsgabor: Won't fix. It does not have to be _a_ response, it can consist of any number of signals. Not to mention that I do not intend to speak about a certain response to a certain message here, but generally about responding to the model.

Issue #523, opened 2017-06-28
The C++ runtime type should be configurable.


nagyandras95:
Currently, the generated c++ code use the threaded runtime as default, even so the runtime use only one thread to execute the model. This thread is seperated from the main thread in this case. The single threaded runtime does not create a seperate thread (so it is a synchronous execution) but can't be configured from the deployment configuration. (only from the code)

PR #527, opened 2017-07-05
Cpp Runtime Type config option

nagyandras95:
This PR 
- Fixes Single runtime, refactoring the statemachine interface and message processing
- Add a new annotation to Configuration description so the generated executor type can be configured.(#523)

### DIFF
--- a/dev/plugins/hu.elte.txtuml.api.deployment/src/hu/elte/txtuml/api/deployment/Configuration.java
+++ b/dev/plugins/hu.elte.txtuml.api.deployment/src/hu/elte/txtuml/api/deployment/Configuration.java
@@ -9,6 +9,7 @@ package hu.elte.txtuml.api.deployment;
  * <pre>
  * <code>@Group(contains = {A.class,B.class}, max = 10, constant = 5, gradient = 0.1)</code>
  * <code>@Group(contains = {C.class})</code>
+ * <code>@Runtime(RuntimeType.THREADED)
  * public class C extends Configuration {}
  * </pre>
  */
--- a/dev/plugins/hu.elte.txtuml.api.deployment/src/hu/elte/txtuml/api/deployment/Group.java
+++ b/dev/plugins/hu.elte.txtuml.api.deployment/src/hu/elte/txtuml/api/deployment/Group.java
@@ -34,4 +34,5 @@ public @interface Group {
 	int constant() default 1;
 
 	int max() default 1;
+	
 }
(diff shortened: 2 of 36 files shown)


djnemeth:
@nagyandras95, for posterity's sake, could you please write a short comment about why you closed this PR?

djnemeth:
Additionally, the generated C++ code doesn't seem to compile. I have attached the [build log](https://github.com/ELTE-Soft/txtUML/files/1166823/build_errors.txt) for the _XMachine1_ model.

nagyandras95:
The XMachine1 should compile. I also had the same error, but it has been fixed the last commit.

djnemeth:
@nagyandras95, you're right, sorry, I was mistakenly testing against an outdated working copy missing f00640d. Now the demo models compile, however, I have found some interesting differences compared to _master_:

1. Demos _machine1_ and _xmachine1_ are apparently missing console output. When executing the main function, only the line contained directly in it ([_Machine and users are starting._](https://github.com/ELTE-Soft/txtUML/blob/b2c1553ee210508c9d5a5fb66bf7f340e326d5f1/examples/demo/machine/src/machine1/j/cpp/main.cpp#L24)) is printed before termination. I believe this might be related to the fact that these models now use the single threaded runtime.
1. It seems the behavior of _xmachine2_ has changed slightly. This might be true for _machine2_ as well which I haven't tested.
   - Java tester, current branch: [xmachine2_java_current.txt](https://github.com/ELTE-Soft/txtUML/files/1170842/xmachine2_java_current.txt)
   - C++ tester, _master_: [xmachine2_cpp_master.txt](https://github.com/ELTE-Soft/txtUML/files/1170843/xmachine2_cpp_master.txt)
   - C++ tester, current branch: [xmachine2_cpp_current.txt](https://github.com/ELTE-Soft/txtUML/files/1170844/xmachine2_cpp_current.txt)
   - C++ tester, _master_ and current diff: https://www.diffchecker.com/xHIrqFCb
1. I believe in the _train_ model the name of initial transitions are ignored, and _Initialize_ is used uniformly instead.
   - C++ tester, _master_: [train_cpp_master.txt](https://github.com/ELTE-Soft/txtUML/files/1170909/train_cpp_master.txt)
   - C++ tester, current branch: [train_cpp_current.txt](https://github.com/ELTE-Soft/txtUML/files/1170911/train_cpp_current.txt)
   - C++ tester, _master_ and current diff: https://www.diffchecker.com/qvRLROWJ

Unfortunately, I cannot guarantee I haven't made the same mistake again, as some of the attached logs were created some days ago and I haven't had the capacity to recheck everything again. Could you please take a look?

PR #530, opened 2017-09-24
Configurable FMU association ends

djnemeth:
In the current implementation, for each FMU class an association has to be defined between the class and the FMU environment. As the latter is a special entity, we generate its C++ implementation separately. Due to the aforementioned association – and by the design of the C++ runtime –, we have to include the association end of the FMU class as an attribute in the generated C++ code of the FMU environment.

Previously, the name of this association end was fixed to `LanderWorld.lander`. This pull request makes it configurable with the `@FMUAssociationEnd` annotation, and also incorporates critical fixes of #529. For additional details, see the (full) commit messages.


djnemeth:
@nboldi, please review.

nboldi:
I think there are unrelated changes in this pull request. (Not related to UML association ends.) Those changes that are related seems good. If you think the rest is OK, I think it can be merged.

djnemeth:
The unrelated changes are from #529 (see the description), which has already been merged to master (in fact, those commits appeared here due to a fast-forward merge from master), so they're OK. Let's merge then! :shipit:

PR #531, opened 2017-10-03
Introduce new base classes in the C++ runtime

nagyandras95:
This PR contains:

- ~ModelClass UML export~ [see the discussion]
- ~ModelClass cpp export in different ways~ [see the discussion] (separate state machine owner and simple class)
- A bigger refactoring in header templates according to new base classes


nboldi:
I will check in the weekend.

dev/plugins/hu.elte.txtuml.stdlib/src/hu/elte/txtuml/stdlib/stdlib.uml
```diff
@@ -547,123 +547,124 @@
...
+      <ownedParameter xmi:id="_nJLtmmg8EeaTWu89rekDRA" name="right">
         <type xmi:type="uml:PrimitiveType" href="pathmap://UML_LIBRARIES/UMLPrimitiveTypes.library.uml#Real"/>
       </ownedParameter>
-      <ownedParameter xmi:type="uml:Parameter" xmi:id="_nJLtm2g8EeaTWu89rekDRA" name="return" direction="return">
+      <ownedParameter xmi:id="_nJLtm2g8EeaTWu89rekDRA" name="return" direction="return">
         <type xmi:type="uml:PrimitiveType" href="pathmap://UML_LIBRARIES/UMLPrimitiveTypes.library.uml#Boolean"/>
       </ownedParameter>
     </ownedOperation>
   </packagedElement>
+  <packagedElement xmi:type="uml:Class" xmi:id="_C8ZIwJYAEeeqkdBjqBM2Qg" name="hu.elte.txtuml.api.model.ModelClass"/>
```
djnemeth: It seems this was the only intended change. Was the mass removal of the `xmi:type` attributes intentional, or was it caused by a graphical editor? If the latter, is this version semantically equivalent with the previous one (apart from the new `packagedElement`)?

nagyandras95: It cased by the graphical editor so I think this changes should be revetred.

djnemeth:
As I can see, polimorf associations (i.e. where at least one of the association ends is a `ModelClass`) are not exported/generated properly. The built-in Ecore editor considers such exported associations invalid (due to an unresolved proxy) and the generated C++ code doesn't compile (`ModelClass` was not declared).

@nboldi, are polimorf associations valid in UML2? If not, how should we export them? Are they really a necessity? We have to consider visualization as well.

nboldi:
My idea was to be able to create generic interfaces. Lets have two components in our domain. They may be connected directly to each other, or they may run in a simulated environment where they communicate through an FMUEnvironment. So I designed their associations to be generic.

I don't really see why associations could not be polymorphic. Since we may have classes `A`, `B` and `C` such as `B` and `C` are extending `A` and we may have an association with one end having type `A`. So in this case association ends are already polymorphic. If the problem is with `ModelClass` alone, we should have it as a class defined in the standard library.

If that is not feasible, I can redesign the example model, but not having the base class available is like programming in Java without being able to write `Object`.

nagyandras95:
I think it would be a good idea to intruduce the ModelClass to the export but is raise some problems in case of UML export too. As I saw If we have a pre-defined class in standard library we could't add properties to it but UML export tries to add the associtaion end property to the owner class  when an end is navigable. We are able to add the Property for Association to the exported UML model remains valid but I think it is not exactly what we want. The C++ association handling should be modified in any case which would be a better task. The PR does not contain regression errors but contains some useful chages so I think should be merged and figured up the next step.

Issue #539, opened 2017-10-31
XtxtUML Collection Framework


kovacsgabor:
Implement the support for the new Collection Framework in X.

See #253 for details.

Issue #197, opened 2015-12-11
Allow collections as attributes and operation parameters


devaigergely81:
The plan is to indicate multiplicities of attributes and operation parameters using collections. The discussion on the different collection types is in a separate issue: #59. Currently we only have the `Collection` collection, but we could already try to allow it as attribute and parameter type.
I see three TODOs related to this:
- Change the corresponding validation rule.
- Check if the runtime works correctly. Do you expect any problem, @kovacsgabor?
- Change the UML export to handle this by using `*` multiplicity in the generated model.

kovacsgabor:
> Check if the runtime works correctly. Do you expect any problem, kovacsgabor?

No, I do not. The API already has a ```Collection.Empty``` class which can be instantiated by the user. It is an empty collection which can be extended with its appropriate methods. (Note, however, that as all txtUML collections, ```Collection.Empty``` is immutable and therefore may only be used to create new collection instances.)

kovacsgabor:
After #542, this issue remains a purely validation problem.

PR #542, opened 2017-11-02
JtxtUML Collection Framework

### LINKED ISSUE #59: Collections: ordering and uniqueness
Currently, an own `Collection` interface is provided in JtxtUML to contain model objects. Every `assoc` call returns an instance of this interface.

To have a more general solution, the following three collection types could be implemented in JtxtUML, based on rAlf:
- Bag (non-unique, unordered)
- Set (unique, unordered)
- Sequence (non-unique, ordered)

Following this change, all `assoc` calls would return either one of these three collections or a single object (possibly `null`), based on the type of the association end that was queried.

### LINKED ISSUE #154: Different collections in case of 0..1 and 1..1 multiplicities?
During talks about #105 on our meetings, another proposal came up to introduce two single-item collections, one to contain 0..1 objects and another one to contain exactly 1. This could improve type safety, as in the latter case, checking the emptiness of the collection is unnecessary, whereas it is required in the former case.

### LINKED ISSUE #253: txtUML Collection Framework
JtxtUML (and in the future, XtxtUML) needs a general support for collections. This requirement and the different proposals about it were already discussed on our meetings multiple times. The following open issues are all connected to this topic: #59, #154, #197.

To sum up, we aim to achieve the following:
- collections to contain model objects, primitive types and data types;
- attribute (field) multiplicities in both model classes and data types;
- uniqueness and ordering;
- easy-to-use syntax which looks as much the same in the different cases (association ends, attribute multiplicities, using collections in action code) as possible;
- `assoc` method on collections.

### LINKED ISSUE #368: Collection.selectAny should return null or throw exception in case of being empty?

kovacsgabor:
Resolves #253 with the exception of the X part (see #539)
Also:
 - resolves #59 
 - resolves #154 
 - resolves #368 

Yet to resolve:
 - #197 (only a validation problem after this pr)
 - #539 (X part of the new collection framework)

## Non-compatible changes in the API

Here comes a detailed list of all changes in the API that may result in an error in the users' code.

### Modeling API
 - Instead of the old `Collection` class, one must use the new Collection Framework. Its base type is the `GeneralCollection` class (as `Collection` is now the base type for unordered, non-unique collections and has 2 type parameters instead of one), it has to be used instead of `Collection` when a general base type is needed. But inside models, even that should not appear, use `Any` instead, the one without restrictions (0..*, unordered, non-unique), and take advantage of the new conversion methods to convert other collections to `Any`.
 - The `Many`, `Some`, `MaybeOne`, `One` and `Multiple` association end types (as well as their `Hidden` counterparts) have been removed. Use `End` or `HiddenEnd` instead, parameterized with a collection type:

```java
class A_B extends Association {
	class a extends End<Any<A>> {}
	class b extends HiddenEnd<One<B>> {}
}
```
 - `GeneralCollection` has a `one()` method instead of `selectAny()` and a `size()` instead of `count()`. It has no `selectAll(<<predicate>>)` method or anything similar.
 - Using custom multiplicities in the model is now only available after defining a new collection type (which is very easy with the help of the existing base types). This, however, is currently not supported in XtxtUML and therefore models written in X will (temporarily) loose their ability to use custom multiplicities.
 - `api.model.assocends` package with all of its contents removed. The `api.model.runtime.collections` package also. All their features can be accessed now through the new Collection Framework.


djnemeth:
Jenkins, retest this, please.

kovacsgabor:
retest this please

kovacsgabor:
ok to test

kovacsgabor:
Jenkins, retest this, please.

djnemeth:
Jenkins, this is ok to test.

nagyandras95:
retest this please

djnemeth:
Jenkins, retest this, please.

kovacsgabor:
Can I ask for a review, @djnemeth? Or someone you suggest.

djnemeth:
Jenkins, retest this, please.

djnemeth:
Jenkins, retest this, please.

PR #546, opened 2017-11-14
Fix lenient execution mode

### LINKED ISSUE #538: Lenient execution mode is not working properly
If the execution mode of a sequence diagram is lenient the result of model testing always says, that the pattern is bigger then the model.

jtamas96:
Resolves #538


kovacsgabor:
ok to test

djnemeth:
Do not merge this before #555. After that we have to check whether there are any applicable changes left from this pulls request.

kmozsi:
As we have discussed at the last meeting, #555 resolves this bug, therefore I close this pull request.

PR #551, opened 2017-11-27
Export mixed models

### LINKED ISSUE #550: The mixed (also containes J and X files) models cannot be exported properly

nagyandras95:
Resolve #550


nagyandras95:
The tests has been reverted since it needed generated code.

djnemeth:
In this case I recommend to add these tests regardless. For a cleaner history, please do this in a separate commit. I have amended the title of your last one to omit the part mentioning tests.

PR #515, opened 2017-04-28
Add microwave example with graphical interface

DianaSuasnavas:
The example is a microwave using graphical user interface, and solve the issue #494

PR #557, opened 2018-01-15
Microwave with gui

### LINKED ISSUE #494: Refresh the microwave oven example model with own images
Whoever takes this task, I will provide her/him with the original model and possibly with images.

kovacsgabor:
An updated version of #515. Resolves #494.

This pull request is based on #555 and #542, it already contains the commits of #555 and #542.


djnemeth:
Do not merge this before #555.

kovacsgabor:
retest this please

kovacsgabor:
retest this please

kovacsgabor:
retest this please

djnemeth:
Visualization of the class diagram silently fails because both Papyrus and JavaScript wizards try to guess – using duplicated code – the underlying model of the layout description from its first annotation, which is `Spacing` here, containing no model information. The lesson here is that wizards should
1) not fail silently,
2) be able to detect the underlying model even in cases like above, and
3) be refactored to eliminate code duplication.

As far as I know wizards are being improved currently on a separate branch, so I suggest a simple reordering here.

Additionally, for some reason `SimpleClockSMDiagram` and `SmartClockSMDiagram` are visualized correctly with JavaScript, but with Papyrus they are displayed as empty diagrams labeled _StateMachine1_ and _StateMachine2_, respectively. @kovacsgabor, can you reproduce this?

djnemeth:
@kovacsgabor has just told me that currently he can't access a developer environment to reproduce what I mentioned above. @szokolai-mate, @jtamas96, could you help us out? See the last paragraph of my previous comment.

djnemeth:
@szokolai-mate, @jtamas96, any update on my previous request?

jtamas96:
If I change StateMachineDiagram<Clock> to StateMachineDiagram<SmartClock> or StateMachineDiagram<SimpleClock> the diagrams are generated properly. Both SmartClock and SimpleClock extends Clock.

djnemeth:
Nice catch! I don't think this is an error, as we want to visualize the state machines of the subclasses indeed, not the one that belongs to their super. Actually, I'm quite surprised that the JS visualization gets it right even with the `Clock` parameter.

PR #579, opened 2018-04-03
Sequence diagram validation

### LINKED ISSUE #559: Sequence diagram validation
Sequence diagrams should be validated during their description (for example blocks without `Sequence.send` calls should not be allowed; lifeline positions must be also validated). 
Please check the already existing solutions for validation.

kmozsi:
This pull request contains validations for sequence diagram descriptions such as:
* the parameter of a lifeline position should be correct,
* if a lifeline is annotated with `Position`, it should be a `ModelClass`,
* each fragment should contain at least one `Sequence.send` or `Sequence.fromActor` call (took private method calls into consideration), 
* calling `Action` methods from `run()` method is not allowed.

Also added tests for these cases and did some refacts in the test code.

Resolves #559.


kmozsi:
Do not merge this before #564.

djnemeth:
Jenkins, retest this, please.

dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/sequencediagram/visitors/Utils.java
```diff
@@ -0,0 +1,169 @@
...
+				collector.report(SequenceErrors.SEND_EXPECTED.create(collector.getSourceInfo(), node));
+			}
+			notSequenceInvocations.forEach(inv -> validateInvokedMethod(inv, visitor, collector));
+		}
+	}
+
+	public static void checkInvalidActionCall(MethodInvocation elem, ProblemCollector collector) {
+		boolean isActionCall;
+		try {
+			isActionCall = elem.resolveMethodBinding().getDeclaringClass().getQualifiedName()
```
djnemeth: This catches only direct action calls, i.e. invoking `void f() { Action.log(""); }` is considered valid but it's clearly not.

dev/tests/hu.elte.txtuml.validation.test/src/hu/elte/txtuml/validation/SequenceDiagramTest.java
```diff
@@ -0,0 +1,82 @@
...
+	}
+
+	private void assertErrors(String fileName, ValidationErrors... errorTypes) throws IOException {
+		CompilationUnit compilationUnit = prepareAST(fileName);
+		compilationUnit.accept(new SequenceDiagramVisitor(mockCollector));
+		Arrays.asList(errorTypes).forEach(errorType -> verify(mockCollector).report(is(errorType)));
+		checkNoOtherErrorRaised();
+	}
+
+	private static ValidationProblem is(ValidationErrors type) {
```
djnemeth: This method is basically a duplication of the one defined for `ModelErrors`. I believe we could eliminate this using appropriate abstractions.

dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/sequencediagram/SequenceDiagramCompilationParticipant.java
```diff
@@ -0,0 +1,26 @@
...
+	public static final String MARKER_TYPE = "hu.elte.txtuml.validation.sequencediagrammarker"; //$NON-NLS-1$
+
+	@Override
+	protected String getMarkerType() {
+		return MARKER_TYPE;
+	}
+
+	@Override
+	protected void validate(CompilationUnit unit, ProblemCollector collector) {
+		if (Utils.isSequenceDiagramDescription(unit)) {
```
djnemeth: I don't think this is necessary here, as the visitor already filters sequence classes. Note that in the case of model errors, the corresponding `isModelElement` check is actually needed, because there we have to look for invalid type declarations inside model packages as well.

dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/sequencediagram/visitors/Utils.java
kmozsi: Dealing with this issue is not trivial at all. I think this solution would do for now as this case is handled with an exception during runtime. Finding a better solution is going to be a future work.

djnemeth: I agree.

dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/sequencediagram/visitors/Utils.java
```diff
@@ -81,13 +94,51 @@ public static void checkSendExists(ProblemCollector collector, Block node) {
...
+			ASTParser parser = ASTParser.newParser(AST.JLS8);
+			parser.setKind(ASTParser.K_COMPILATION_UNIT);
+			parser.setSource(unit);
+			parser.setResolveBindings(true);
+			CompilationUnit cu = (CompilationUnit) parser.createAST(null);
+
+			MethodDeclaration decl = (MethodDeclaration) cu.findDeclaringNode(binding.getKey());
+			decl.getBody().accept(visitor);
+		} catch (NullPointerException ex) {
+			Logger.sys.debug("Method calling cannot be validated.");
```
djnemeth: Should be _call_.

Issue #560, opened 2018-02-03
Sequence diagram visualization and execution


kmozsi:
Currently, sequence diagram visualization and execution are not so user-friendly, furthermore visualizing test diagrams mixed with model diagrams is confusing. Thus it would be good to separate these features. 

For test execution the user has to write code, which can be avoided by providing a UI feature with a basic executor implementation. A possible way is to use the package/project explorer context menu of a sequence diagram (e.g. 'execute tests'), execute the user code, and finally indicate the errors on the UI.

As similar dialogs are already implemented, some refact might be done.

This improvement could also fix #536.

kmozsi:
Since the related pull requests have been merged, I close this issue.

Issue #81, opened 2015-10-25
Split txtUML into separate features


devaigergely81:
Proposal for the features (all in the txtUML category):
1: JtxtUML (Java APIs)
2: XtxtUML (DSL syntax) -- dependency: 1
3: EMF-UML2 model export -- dependency: 1
4: Papyrus diagram generation -- dependency: 3
5: State machine animation -- dependency: 4
6: C++ code generation -- dependency: 3

kovacsgabor:
1 would consist of:
hu.elte.txtuml.api.diagnostics
hu.elte.txtuml.api.layout
hu.elte.txtuml.api.model
hu.elte.txtuml.api.stdlib
hu.elte.txtuml.utils

kovacsgabor:
2 would consist of:
hu.elte.txtuml.xtxtuml
hu.elte.txtuml.xtxtuml.lib
hu.elte.txtuml.xtxtuml.ui

kovacsgabor:
3 would consist of:
hu.elte.txtuml.export.uml2

kovacsgabor:
4 would consist of:
hu.elte.txtuml.export
hu.elte.txtuml.export.papyrus
hu.elte.txtuml.layout.export
hu.elte.txtuml.layout.visualizer
hu.elte.txtuml.stdlib

kovacsgabor:
5 would consist of:
hu.elte.txtuml.diagnostics

kovacsgabor:
6 would consist of:
hu.elte.txtuml.export.cpp

PR #581, opened 2018-04-12
Sequence diagram execution

jtamas96:
Related to #560.

Contains:

- Sequence diagram execution from ui.

- Execution can be cancelled by the user.

- Result of sequence diagram execution is written to console.


djnemeth:
> I think these UI-related changes do not belong to `api.model.execution`.

Agreed.

> If we cannot find a better location, I propose `export.plantuml` instead.

I don't think there's a better location. We might consider renaming the project.

Additionally, instead of writing directly to the console, couldn't we use our [_Logger_](https://github.com/ELTE-Soft/txtUML/blob/94c2efaf648f9b3a196fc334411946ef3949f1e4/dev/plugins/hu.elte.txtuml.utils/src/hu/elte/txtuml/utils/Logger.java#L6)?

jtamas96:
> Additionally, instead of writing directly to the console, couldn't we use our Logger?

The idea is to give the user a direct response about the results of the last run. I think the console view is a better option for this purpose because we can isolate the results in their own console.

> We might consider renaming the project.

Should I name it to seqDiag (or something like that), or just move the solution and leave the naming as it is for now?

djnemeth:
I suggest the name `seqdiag`. However, with @szokolai-mate we're considering moving Eclipse-dependent control code from existing modules to a dedicated project. I believe this would apply here as well, but I'm generally concerned about separating process control logic from their respective components. @jtamas96 and @kmozsi, what do you think?

jtamas96:
I think in some cases moving the Eclipse-dependent control code would be very useful. For example currently 4 projects are related to the txtUML menu, so we have to make a menu contribution extension in each of the 4 projects in order to add a command to our txtUML menu. I think this is not so readable and I couldn't find a way to determine the order of the commands in the menu, but defining the whole menu with it's commands in one project would make the ordering possible.

kmozsi:
> I suggest the name `seqdiag`.

I like it. 👍 

>  I'm generally concerned about separating process control logic from their respective components.

As process control logic does not strongly belong to its current context, I think the refact is a good idea, and also (eclipse) dependencies would be rationalized by doing so. 

For now, I think the project should be renamed to `seqdiag`, and please move your changes there. Make sure to separate these improvements well from PlantUML export.

kmozsi:
Jenkins, retest this, please.

dev/plugins/hu.elte.txtuml.api.model.execution/build.properties
```diff
@@ -1,4 +1,5 @@
 source.. = src/
 output.. = bin/
 bin.includes = META-INF/,\
-               .
+               .,\
+               plugin.xml
```
kmozsi: Please also remove this.

djnemeth:
> As process control logic does not strongly belong to its current context, I think the refact is a good idea, and also (eclipse) dependencies would be rationalized by doing so.

On the other hand, centralizing control logic in a separate project would make the whole feature less modular, i.e. we would lose the ability to achieve #81. @kmozsi, @szokolai-mate, @jtamas96, what do you think?

djnemeth:
> On the other hand, centralizing control logic in a separate project would make the whole feature less modular, i.e. we would lose the ability to achieve #81. @kmozsi, @szokolai-mate, @jtamas96, what do you think?

I have created a dedicated issue for this problem in the form of #601. Not to be dealt with in 0.7.0.

> 8771a56 Fix context menu

@kmozsi, please don't forget to review these changes as well.

kmozsi:
Jenkins, retest this, please.

PR #609, opened 2018-11-06
Improve JtxtUML validator markers

### LINKED ISSUE #506: JtxtUML validator markers are incorrectly placed
Some of the markers generated by the JtxtUML validator are placed incorrectly.

1. Marking whole source elements (including whitespaces), not only identifiers:
whole_source_element

2. Marking one character less than needed:
too_short
too_short2

These issues have to be addressed by placing markers exactly on the most relevant source segment with regards to the given validation problem.

berpeti:
Fixes #506.

The issue was that our validator placed error markers incorrectly on elements like classes and operations. This pull request fixes these problems by separating the responsibility of detecting and displaying errors.

The proposed separation takes place between the visitors which detect errors (e.g. `ModelClassVisitor`), and the abstract classes that represent error markers (`AbstractValidationProblem`, `ModelValidationError`, `SequenceValidationError`). The responsibility of the detectors is to _select and pass_ exactly the faulty `ASTNode` to a newly created error instance (e.g. if the return type of an operation is incorrect, this reference node will be passed), while error classes determine how to _place_ the error marker on the given faulty node (e.g. when the incorrect node is a class definition, only the name of the class will be marked).

The following before-after examples demonstrate how the proposed changes improve the placement of JtxtUML error markers.

| Before | After |
| -------|-------|
| bad1 | good1 |
| bad2 | good2 |

### DIFF
--- a/dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/common/AbstractValidationProblem.java
+++ b/dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/common/AbstractValidationProblem.java
@@ -5,7 +5,6 @@ import java.util.List;
 import org.eclipse.jdt.core.compiler.CategorizedProblem;
 import org.eclipse.jdt.core.dom.ASTNode;
 import org.eclipse.jdt.core.dom.MethodDeclaration;
-import org.eclipse.jdt.core.dom.PrefixExpression;
 import org.eclipse.jdt.core.dom.SingleVariableDeclaration;
 
 import hu.elte.txtuml.validation.model.ModelErrors;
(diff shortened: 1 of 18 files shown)


djnemeth:
Jenkins, retest this, please.

djnemeth:
@berpeti, I've noticed that in 45c1553 you changed some of the error detection code according to the proposed responsibility separation convention. Have you checked whether _all_ visitors satisfy this principle?  (i.e. there are no calls like `.getName()` at places of error detection)

berpeti:
Yes, I checked. All visitors satisfy this principle.

djnemeth:
Jenkins, retest this, please.

dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/common/AbstractValidationProblem.java
```diff
@@ -12,14 +12,14 @@
...
-	private int sourceStart;
-	private int sourceEnd;
+	protected int sourceStart;
+	protected int sourceEnd;
 	private int lineNumber;
 
 	public AbstractValidationProblem(SourceInfo sourceInfo, ASTNode node) {
 		this.sourceInfo = sourceInfo;
 		this.sourceStart = node.getStartPosition();
-		this.sourceEnd = node.getStartPosition() + node.getLength() - 1;
```
djnemeth: Do we know the reason for this?

PR #610, opened 2018-11-06
Fix par fragment

### LINKED ISSUE #593: Sequence diagram execution does not terminate when PAR has extra sends
Sequence diagram execution does not terminate when a PAR fragment contains extra send expectations compared to the model.

jtamas96:
Fixes #593 and extends the unit tests with a sequence diagram which contains extra sends in par fragments and after the par fragments.


djnemeth:
What is the status of this pull request?

kmozsi:
Jenkins, retest this, please.

Issue #605, opened 2018-09-27
Support visualization of hierarchic state machines


szokolai-mate:
Currently, hierarchical state machines are supported by both language syntaxes in the form of composite states. As far as I know, composite states work as intended, and are correctly exported into the UML2 model.

This, however is not displayed neither in Papyrus nor in JointJS diagrams.

@djnemeth Please verify that this is correct, and provide us with the un-merged old code pertaining to this issue.

djnemeth:
> Composite states work as intended, and are correctly exported into the UML2 model. This, however is not displayed neither in Papyrus nor in JointJS diagrams. Please verify that this is correct.

Correct. The current version of the abstract diagram aligner and the Papyrus/JavaScript visualizer in _master_ handles composite states as simple (flat) ones.

> Please provide us with the un-merged old code pertaining to this issue.

The code is available in branch _[new/papyrus-export](https://github.com/ELTE-Soft/txtUML/compare/new/papyrus-export)_. The related issue is #214.

Issue #577, opened 2018-04-03
Improve code quality of visualization-related components


djnemeth:
The following modules have to be revised:
- _hu.elte.txtuml.export.javascript_
- _hu.elte.txtuml.export.papyrus_
- _hu.elte.txtuml.layout.export_
- _hu.elte.txtuml.layout.visualizer_

Possible improvements:
- eliminate code duplication
- remove unnecessary dependencies
- find better abstractions
- increase separation

This is to make future developments (e.g. #576, #534 and possibly #214) easier.

djnemeth:
I recommend to address this in branch _[visualization-refactoring](https://github.com/ELTE-Soft/txtUML/tree/visualization-refactoring)_.

PR #574, opened 2018-03-29
Sequence diagram visualization

jtamas96:
Resolves #536, and related to #560

Contains:

- New dialog for sequence diagram visualization.

- Separated PreferencesManager from Papyrus export project.

- PlantUML export project no longer depends on Papyrus export project.

Issue #524, opened 2017-07-01
Papyrus diagram generation fails if the uml2 model already exists


kovacsgabor:
If the uml2 model already exists, the Papyrus diagram generation starts but cannot finish properly (sometimes an error is raised, sometimes it results in a corrupt model).

To generate a Papyrus diagram, I had to manually delete the existing model.

djnemeth:
I could reproduce this issue on the master branch using a runtime Eclipse. I generated diagrams from the clock demo, and when I tried to re-visualize it once again, error messages appeared and the diagram became corrupted. See the [stack trace](https://github.com/ELTE-Soft/txtUML/files/1123670/revisualization-trace.txt). JointJS visualization also seems affected.

PR #614, opened 2018-11-24
Hierarchical State-Machine visualization

### LINKED ISSUE #582: Improve visualization wizards
If a diagram description starts with a [`Spacing`](https://github.com/ELTE-Soft/txtUML/blob/1739c05df8478b6f35e4cdbb97cfbe2e2154220d/dev/plugins/hu.elte.txtuml.api.layout/src/hu/elte/txtuml/api/layout/Spacing.java#L14) statement, both Papyrus and JS visualization silently fails because their wizards try to guess – using duplicated code [[1](https://github.com/ELTE-Soft/txtUML/blob/1739c05df8478b6f35e4cdbb97cfbe2e2154220d/dev/plugins/hu.elte.txtuml.export.papyrus/src/hu/elte/txtuml/export/papyrus/wizardz/TxtUMLVisualizeWizard.java#L95), [2](https://github.com/ELTE-Soft/txtUML/blob/1739c05df8478b6f35e4cdbb97cfbe2e2154220d/dev/plugins/hu.elte.txtuml.export.javascript/src/hu/elte/txtuml/export/javascript/wizardz/TxtUMLVisualizeWizard.java#L88)] – the underlying model of the layout description from its first annotation, but in case of `Spacing` this is not possible, as it doesn't contain the necessary model information to do so.

The lesson is that wizards should
  1. not fail silently,
  2. be able to detect the underlying model even in cases like above, and
  3. be refactored to eliminate code duplication.

### LINKED ISSUE #585: Problems during re-visualization
If one tries to re-visualize a model for which there is an already existing Papyrus diagram in the project's _gen_ folder, the visualization fails. If the corresponding diagram is opened during re-visualization, a `ConcurrentModificationException` occurs and the _gen_ folder has to be cleaned. Otherwise the process seemingly succeeds, but in reality, only the _.di_ and _.notation_ files disappear from the _gen_ folder. In this case, however, a second re-visualization attempt actually creates and opens the correct diagram. This is only for Papyrus – JS works as intended, but it could be improved too.

It would be great if
1. Papyrus diagrams could be re-visualized without the need of cleaning _gen_, whether the previously generated diagram is opened or not – here we could simply overwrite the existing diagram;
2. the user could decide whether they wants to overwrite the existing diagram or cancel the visualization – this is for both Papyrus and JS.

It would be even greater if we could append "tabs" to existing diagrams. For example, if one visualizes a class diagram from a model in Papyrus and then decides to add a state machine diagram from the same model, we wouldn't need to replace the whole _.di_ file, but simply append the new diagram to the existing one. The same for JS. As this doesn't look like a trivial task, I'm willing to create a separate issue for it.

szokolai-mate:
Main issue: #605

Includes the branch [new/papyrus-export](https://github.com/ELTE-Soft/txtUML/tree/new/papyrus-export), implementing all of its changes and functionality.
These include:

- proper exportation and arranging of hierarchical state-machines
- proper visualization of hierarchical state-machines in Papyrus
- a new element of the diagram language: `Box`
    - an empty `Box` is eqvivalent to the old `Phantom`
    - a `Box` can also have a sub-diagram with the `@Inside` annotation. This is the way to make hierarchical state-machines.
- refactoring of visualization components: wizards, visualization and Papyrus components
    - new project `export.diagrams.common` now properly separates common elements in visualization components

Main issue: #214 
Resolves #585, #577, #524, #443, #516

Additional work on this branch:

A separate refactoring of the wizards and Papyrus components was already present on `master` from #574. Consolidated the two refactorings, with the changes from `new/papyrus-export` taking priority. Kept now unused classes in `export.papyrus` for future use.

As this PR makes the branch `visualization-refactoring` more or less obsolete, I've included the fix for #582 in this branch aswell. Resolves #582 .

JointJS visualization now produces identical hierarchical diagrams to Papyrus, handles the new `Box` diagram element properly, has proper progress bar and is cancelable at any time.

### DIFF
--- a/dev/releng/hu.elte.txtuml.parent/pom.xml
+++ b/dev/releng/hu.elte.txtuml.parent/pom.xml
@@ -268,6 +268,36 @@
 							</configuration>
 						</execution>
 					</executions>
+					<dependencies>
+						<!-- make EMF v2.12 dependencies explicit, see
+							 https://github.com/eclipse/xtext/issues/1231#issuecomment-421418880 -->
+						<dependency>
+							<groupId>org.eclipse.emf</groupId>
+							<artifactId>org.eclipse.emf.common</artifactId>
+							<version>2.12.0</version>
+						</dependency>
+						<dependency>
+							<groupId>org.eclipse.emf</groupId>
+							<artifactId>org.eclipse.emf.ecore</artifactId>
+							<version>2.12.0</version>
+						</dependency>
+						<dependency>
+							<groupId>org.eclipse.emf</groupId>
+							<artifactId>org.eclipse.emf.ecore.xmi</artifactId>
+							<version>2.12.0</version>
+						</dependency>
+						<dependency>
+							<groupId>org.eclipse.emf</groupId>
+							<artifactId>org.eclipse.emf.codegen</artifactId>
+							<version>2.11.0</version>
+						</dependency>
+						<dependency>
+							<groupId>org.eclipse.emf</groupId>
+							<artifactId>org.eclipse.emf.codegen.ecore</artifactId>
+							<version>2.12.0</version>
+						</dependency>
+						<!-- end of explicit EMF v2.12 dependencies -->
+					</dependencies>
 				</plugin>
 
 				<!--This plugin's configuration is used to store Eclipse m2e settings only. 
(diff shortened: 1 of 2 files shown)


djnemeth:
Jenkins, retest this, please.

djnemeth:
Jenkins, retest this, please.

djnemeth:
I postpone this for now.

PR #616, opened 2019-01-27
Rework the XtxtUML execution element

### LINKED ISSUE #430: Rework the 'execution' element of XtxtUML

berpeti:
This pull request is intended to resolve #430.

For this purpose the pull request introduces the new X-style execution element, which is translated to a J-style execution. The motivation for this new X-style element is that for X-style models only using J-style executions was possible.

The structure of these X-style elements is the following:

<pre><b>execution</b> «name» {
   «attributes»   // mapped to J-style execution attributes
   <b>initialization</b> {«code»}   // mapped to J-style initialization()
   <b>configure</b> {«code»}   // used for the J-style configure(Settings s)
   <b>before</b> {«code»}   // mapped to J-style before()
   <b>during</b> {«code»}   // mapped to J-style during()
   <b>after</b> {«code»}   // mapped to J-style after()
}</pre>

It is possible to not use each of the above mentioned methods.

The `configure` block is not mapped directly to J-style `configure`. It's mapped to the `configureExecution` function, and the `name`, `logLevel`, `checkLevel`, `timeMultiplier` attributes. In the X-style configure block values can be set to these attributes. Then the J-style configure block is generated, which calls this function and then sets the values for the `Settings s` parameter with conditional assignments. Besides this, only in this configure block is it possible to "implement" the J-style `name()` function with assigning value to the `name` attribute. From these assignments the above mentioned J-stlye function is generated

This pull request also contains for this language element Eclipse support i. e. syntax highlight, static validation, compilation to JtxtUML, auto formatting, import organizing, quickfixes and content assist. (New code added only for validation, formatting, compilation. The code of the other features were already working.)

This pull request includes beside these unit tests and demos for the new execution element.


djnemeth:
@berpeti, please check my changes in ba792c2.

berpeti:
I checked, everything is working. Thank you for your help.

Issue #615, opened 2018-12-11
C++ deployment configuration groups should be revised.


nagyandras95:
- The number of the threads should depend on the plathrom (number of cores + 1).
- Removing threads requires many resources, the number of threads should be fixed.
- The distribution ratio should be configured in a group instaed of the lienar function.

For intsance:
```
@RuntimeType(Threaded)
@Group(contains = {A.class, B.class}, rate = 0.3)
@Group(contains = {C.class}, rate = 0.7)
class Conf extends Configuration {}
```

djnemeth:
@nagyandras95, please provide a concrete example for this concept. That is, a step-by-step sample execution which shows how threads are managed as model instances are created and deleted.

PR #617, opened 2019-01-31
New c++ config prototype

nagyandras95:
Prototype for #615. 
The threaded runtime works fixed threads depend on the current platform. 
Thre threads dispersion can be configured within groups.


nagyandras95:
retest this please

Issue #608, opened 2018-10-20
Improve sequence diagram validation


kmozsi:
To make definition of _correct_ sequence diagram descriptions easier, two more validation rules could be added. An incomplete solution for both rules has been implemented earlier (see #579).

- Each fragment of `run` method should contain at least one `Sequence.send` or `Sequence.fromActor` call. Invoked (private) methods of `run` should also be checked. This rule has been reverted before, see https://github.com/ELTE-Soft/txtUML/pull/597#pullrequestreview-129403811. 
- Calling `Action` methods from `run` should not be allowed due to threading issues. Currently only direct `Action` calls are handled, see https://github.com/ELTE-Soft/txtUML/pull/579#discussion_r195929821.

jtamas96:
A part of this issue was solved in #618. The unsolved parts were moved into #623, #624 and #621, so I am closing this issue.

PR #618, opened 2019-02-19
Improve sequence diagram validation

jtamas96:
This pull request solves a part of #608.
Validation of the following are implemented:
- The run method of a SequenceDiagram is valid if:
   - it contains at least one fragment and all of the fragments are valid, or
   - it contains at least one send invocation
- A loop fragment is valid if:
   - its body contains at least one fragment and all of the fragments are valid, or
   - it contains at least one send invocation
- An alt/opt fragment is valid if all of its branches are valid.
- An alt/opt fragment branch is valid if:
   - its body contains at least one fragment and all of the fragments are valid, or
   - it contains at least one send invocation
- A par fragment is valid if all of its arguments are valid.

The solution is not complete, the pull request is created, so we can cut the remaining part of the issue into smaller pieces.


kmozsi:
Retest, please.

kmozsi:
Jenkins, retest this, please.

dev/plugins/hu.elte.txtuml.validation/src/hu/elte/txtuml/validation/sequencediagram/messages.properties
```diff
@@ -3,3 +3,4 @@ SendExpected_message=Message sending expected.
 InvalidPosition_message=Lifeline position must not be negative.
 InvalidLifelineDeclaration_message=The declared lifeline must be a ModelClass.
 InvalidActionCall_message=Action code is not allowed in this context.
+CycleDetected_message=There is a possible cycle in the SequenceDiagram.
```
kmozsi: Insert a newline here please.
Also, i think it should be 'sequence diagram' instead of 'SequenceDiagram', as this sentence is not about the SeqDiag class.

PR #620, opened 2019-02-24
Proxy objects

### LINKED ISSUE #590: Expected communication cannot be expressed in case of unknown sender or target
`Sequence.send(sender, signal, target)` cannot be used to express expected communication when a reference of sender or target is not visible from sequence diagram code (e.g. objects created inside a model class). 

A solution might be to introduce dummy objects to substitute unknown senders/targets. 
For example a `Sequence.create(MyModelClass.class)` method would be added to create such a dummy object, which would be wrapped as `Proxy<MyModelClass>`. We would make a distinction between two Proxy objects, as they would represent different model objects, and also any of these Proxy objects would be different than a model object that can be referenced from sequence diagram code.

kmozsi:
Resolves #590.

- Introduced proxy objects to substitute `ModelClass` instances that cannot be referenced from sequence diagram code.
- Declaration of lifelines is made explicit. Only `Lifeline` and `Proxy` fields are visualized on the diagram. 
- Added a test to pingpong demo using proxies.
- Rewrote tests and sequence diagrams of demos.


dev/plugins/hu.elte.txtuml.api.model.execution/src/hu/elte/txtuml/api/model/execution/impl/seqdiag/Message.java
```diff
@@ -4,80 +4,86 @@
...
 		try {
-			return BasedOnFields.equal(getWrapped(), other.getWrapped());
+			areSignalsEqual = BasedOnFields.equal(getWrapped(), expected.getWrapped());
 		} catch (IllegalAccessException e) {
 			return false;
 		}
+
+		if (areSignalsEqual) {
+			if (sender != null && expected.sender != null && !expected.sender.hasParticipant()) {
+				expected.sender.bindTo(sender.getParticipant().get());
```
jtamas96: I think we should check the classes of the senders before binding, because they might not be the same and in this case we have an error.

dev/plugins/hu.elte.txtuml.api.model.execution/src/hu/elte/txtuml/api/model/execution/impl/seqdiag/Message.java
```diff
@@ -4,80 +4,86 @@
...
+			}
+
+			if (!expected.target.hasParticipant()) {
+				expected.target.bindTo(target.getParticipant().get());
```
jtamas96: The same problem as above.

dev/plugins/hu.elte.txtuml.api.model.execution/src/hu/elte/txtuml/api/model/execution/impl/seqdiag/Message.java
kmozsi: I agree, thank you!

dev/plugins/hu.elte.txtuml.api.model.execution/src/hu/elte/txtuml/api/model/execution/impl/seqdiag/Message.java
```diff
@@ -4,80 +4,88 @@
...
+				expected.sender.bindTo(sender.getParticipant().get());
+			}
+
+			if (!expected.target.hasParticipant()
+					&& target.getParticipantType().equals(expected.target.getParticipantType())) {
+				expected.target.bindTo(target.getParticipant().get());
+			}
+		}
+
+		return areSignalsEqual;
```
jtamas96: We should also return false, if the class of the proxy is different on either side, then the expected class. As for the binding, I am not sure if we should bind the target, if the sender can't be bound and vice versa.

jtamas96: The problem above will be examined, and solved in another issue, so I think this can be merged now.

kmozsi: I agree.

Issue #534, opened 2017-10-24
Support state machine instance list in visual debugger


kmozsi:
Supporting multiple instances would be desired, but currently we can debug only one instance of each state machine.

djnemeth:
Due to limitations of Papyrus, the team have decided to support this feature only on JavaScript diagrams. As a result, this now depends on #576.

djnemeth:
Postponed.

szokolai-mate:
Updated description of this issue:

Currently, the JavaScript debug view only supports displaying one state machine diagram at a time.
The DiagnosticsServer already supplies the browser client with runtime information, which consists of all active state machines, thier current active state and their unique ID. We can also extend this with the runtime name of the state machine, given by the user.

Decisions to make:

- How do we display these state machine instances, which are theoratically infinite in number?
    - On different tabs (like seperate state machine models)?
        - This way, we would still only see one state machine at a time.
    - On one tab?
        - How do we merge the SVG canvases?
        - How do we arrange these canvases?
        - Reduce the drawn elements by applying a filter to which state machine instances are displayed?

szokolai-mate:
After a team discussion, we decided on the following:

The "display unit" is one class: we can simultaneously see all instances of **one class**, if we choose to, and we can change between these views by selecting a different state machine diagram.

Features to achieve:

Ability to select which instances we wish to display
Displaying multiple instances
    either by merging the input, or merging the SVG elements generated by JointJS
Parameterizing the generation of the imput for said instances: name and id
Animating the instances: filtering animated state machines by id

PR #631, opened 2019-04-09
State machine instance selection

idontspeakgithub:
Main issue: #534.

This PR adds initial support for state machine instance selecting.

### DIFF
--- a/dev/plugins/hu.elte.txtuml.export.javascript/src/hu/elte/txtuml/export/javascript/resources/res/pollingRefresh.js
+++ b/dev/plugins/hu.elte.txtuml.export.javascript/src/hu/elte/txtuml/export/javascript/resources/res/pollingRefresh.js
@@ -30,9 +30,12 @@ function refreshElements(queryPort){
 
 		if(response.status == 200){
 			var json = JSON.parse(response.responseText);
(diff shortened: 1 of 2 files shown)


kmozsi:
Jenkins, retest this, please.

orsidudas:
Nice work! I have a few notes before I apply:
- It would be good if you exchange the 'obj0, obj1, ...' names the instances' actual name.
- I have to click multiply times on the obj-s before the page responds for my request. It would be good, if the page would respond immediately.
- In the Train example, it looks like the object are reversed. When I'm on the Lamp's object, the web address says it's the engine's obj and vice versa.

idontspeakgithub:
>    It would be good if you exchange the 'obj0, obj1, ...' names the instances' actual name.

This is the most user friendly name I get from the server.

>    I have to click multiply times on the obj-s before the page responds for my request. It would be good, if the page would respond immediately.

I think this issue is out of my scope.

>    In the Train example, it looks like the object are reversed. When I'm on the Lamp's object, the web address says it's the engine's obj and vice versa.

Thanks, fixed.

kmozsi:
Jenkins, retest this, please.

